PR #127, opened 2024-02-13
Top level taxonomy

smendis-scottlogic:
Top level taxonomy categorization.
Send level taxonomy categorization for compute, networking, storage and database done.


eddie-knight (approved):
This is great progress. Some elements can be refined further over time.

AdrianHammond (approved):
This PR was reviewed in Taxonomy WG and it was agreed that we would merge and then take to other WG's for their comments.

PR #135, opened 2024-03-05
Initial version of the object store taxonomy

smendis-scottlogic:


services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+
+| Taxonomy ID | Feature | Description |
+| ----------- | ------- | ----------- |
+| CCC-020101 | Buckets | Concept of having uniquely identifiable containers or buckets to store objects. |
+| CCC-020102 | Metadata | Support storing, accessing, and managing of object metadata for stored objects. |
+| CCC-020103 | Scalability - Capacity Limit | Ability to store unlimited number of objects under a given maximum total capacity. |
+| CCC-020104 | Scalability - Object Size Limit | Ability to store large objects under a given maximum object size. |
+| CCC-020105 | Durability | High durability for stored objects through redundancy and replication. |
+| CCC-020106 | Availability | High availability for stored objects through replication over multiple availability zones within a region. |
+| CCC-020107 | Performance - Transaction Rate Limits | High throughput and low latency for read/write operations under a given maximum transaction rate limits.  |
```
sshiells-scottlogic, line 17: This doesn't read quite right to me. Maybe it could be changed to:

High throughput and low latency for read/write operations under given maximum transaction rate limits.

or

High throughput and low latency for read/write operations under a given maximum transaction rate limit.

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020109 | Storage Classes | Having different storage classes for frequently and infrequently accessed objects. |
+| CCC-020110 | Lifecycle Policies | Ability to define policies to automate data management tasks. |
+| CCC-020111 | Versioning | Ability to keep multiple versions of an object in the same object store (bucket). |
+| CCC-020112 | Compliance and Governance | Ability to create locks on objects disabling modification or/and deletion of an object for a given period of time. |
+| CCC-020113 | Event Notifications | Publish object level events for creation, deletion and modification of objects allowing users to trigger actions in response. |
+| CCC-020114 | Encryption at Rest | Objects are encrypted when storing using encryption keys. |
+| CCC-020115 | Encryption in Transit | Objects are encrypted in transit, using SSL/TSL. |
+| CCC-020116 | Role Based Access Control | Ability to limit the users/roles who can access the object store. |
+| CCC-020117 | Object Based Access Control | Ability to control access to specific objects on the store. |
+| CCC-020118 | Logging | Ability log access allowing the users to track requests made to the object store. |
```
sshiells-scottlogic, line 28: Should this be "users/roles" to be consistent with the terms used in the Role Based Access Control entry?

sshiells-scottlogic (commented):
Looks good, a couple of small comments about possible minor tweaks to the wording.

services/storage/object/taxonomy.md
smendis-scottlogic: Thanks! selected the first version as there are 2 rates, max read limit and max write limit

services/storage/object/taxonomy.md
smendis-scottlogic: Make sense. Updated. Thanks!

AdrianHammond (commented):
Hi @smendis-scottlogic @eddie-knight  - LGTM - thank you Sonali

As a group I think we need to agree on standard for numbering the Taxonomy id's.

sshiells-scottlogic:
> Hi @smendis-scottlogic @eddie-knight - LGTM - thank you Sonali
> 
> As a group I think we need to agree on standard for numbering the Taxonomy id's.

Once a number system is agreed upon I think it would be good to have this through the upper levels so it is easier to navigate. For example (just using the system @smendis-scottlogic  has went with for now):

image

smendis-scottlogic:
Yes I followed the pattern we did in RDMS and was not sure what to put in the place of RDMS. In here I have come up with this numbering system and hoping to present it to the team on the WG meeting. 
first 2 digits - top level taxonomy
second 2 digits - second level taxonomy
last 2 digits - taxonomy id
Open for other suggestions. Thanks!

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020106 | Availability | High availability for stored objects through replication over multiple availability zones within a region. |
+| CCC-020107 | Performance - Transaction Rate Limits | High throughput and low latency for read/write operations under given maximum transaction rate limits.  |
+| CCC-020108 | Performance - Querying | Ability to perform simple select queries to retrieve only a subset of objects from the bucket. |
+| CCC-020109 | Storage Classes | Having different storage classes for frequently and infrequently accessed objects. |
+| CCC-020110 | Lifecycle Policies | Ability to define policies to automate data management tasks. |
+| CCC-020111 | Versioning | Ability to keep multiple versions of an object in the same bucket. |
+| CCC-020112 | Compliance and Governance | Ability to create locks on objects disabling modification or/and deletion of an object for a given period of time. |
+| CCC-020113 | Event Notifications | Publish object level events for creation, deletion and modification of objects allowing users to trigger actions in response. |
+| CCC-020114 | Encryption at Rest | Objects are encrypted when storing using encryption keys. |
```
rgriffiths-scottlogic, line 24: I infer from this that there is the possibility of storing objects without them being encrypted and that encrypted storage is only possible when using encryption keys. Is that the correct way to read this?

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020115 | Encryption in Transit | Objects are encrypted in transit, using SSL/TSL. |
```
rgriffiths-scottlogic, line 25: This appears to be "always on" - is that the case?

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020116 | Identity Based Access Control | Ability to limit the users/roles who can access the bucket. |
```
rgriffiths-scottlogic, line 26: Do we want to use "bucket"? Or would "object store" be more generic? Or something similar, of course!

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020117 | Object Level Access Control | Ability to control access to specific objects on the bucket. |
```
rgriffiths-scottlogic, line 27: Ditto to previous comment. S3 has buckets but Azure has blobs, for example.

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020118 | Logging | Ability log access allowing the users/roles to track requests made to the bucket. |
```
rgriffiths-scottlogic, line 28: "Ability to log" (missing a "to")

rgriffiths-scottlogic (commented):
A few comments (all minor!) from me.

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+## Taxonomy
+
+| Taxonomy ID | Feature | Description |
+| ----------- | ------- | ----------- |
+| CCC-020101 | Buckets | Concept of having uniquely identifiable containers or buckets to store objects. |
+| CCC-020102 | Metadata | Support storing, accessing, and managing of object metadata for stored objects. |
+| CCC-020103 | Scalability - Capacity Limit | Ability to store unlimited number of objects under a given maximum total capacity per bucket. |
+| CCC-020104 | Scalability - Object Size Limit | Ability to store large objects under a given maximum object size. |
+| CCC-020105 | Durability | High durability for stored objects through redundancy and replication. |
+| CCC-020106 | Availability | High availability for stored objects through replication over multiple availability zones within a region. |
```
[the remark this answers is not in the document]
smendis-scottlogic: Both AWS and Azure use the terminology of region and availability zone. GCP refer to it as region and zone. But the concept is same among all cloud providers. So I suggest we keep it as it is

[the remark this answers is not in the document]
sshiells-scottlogic: We could put it as:

... replication over multiple (availability) zones within a region.

?

It kind of covers all basis. I am not sure it makes too much of a difference though.

services/storage/object/taxonomy.md
```diff
@@ -0,0 +1,28 @@
...
+| CCC-020107 | Performance - Transaction Rate Limits | High throughput and low latency for read/write operations under given maximum transaction rate limits.  |
+| CCC-020108 | Performance - Querying | Ability to perform simple select queries to retrieve only a subset of objects from the bucket. |
+| CCC-020109 | Storage Classes | Having different storage classes for frequently and infrequently accessed objects. |
+| CCC-020110 | Lifecycle Policies | Ability to define policies to automate data management tasks. |
+| CCC-020111 | Versioning | Ability to keep multiple versions of an object in the same bucket. |
+| CCC-020112 | Compliance and Governance | Ability to create locks on objects disabling modification or/and deletion of an object for a given period of time. |
+| CCC-020113 | Event Notifications | Publish object level events for creation, deletion and modification of objects allowing users to trigger actions in response. |
```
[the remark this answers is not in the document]
smendis-scottlogic: S3, Azure Blob Storage does not support read events natively as of my knowledge. I think it is important to respond when an object got created/modified or deleted other that when read. You can have access logs to track all requests to the bucket and have ACL to control over who can read.

services/storage/object/taxonomy.md
smendis-scottlogic: Bucket is again a core concept in cloud storage base on my research just like the region and availability zone and well known. Will change if others also thinks it is better to use the term object store instead of the term bucket.

services/storage/object/taxonomy.md
smendis-scottlogic: Best practice is to use https. But can use http urls.

eddie-knight:
@AdrianHammond I'd like feedback on this review process. Here are my thoughts... building toward a suggestion on the final point.

1. We are in a phase where we should be moving in quick iterations, so we shouldn't hamper this down with a long review process.
2. We don't have a process right now to determine what the blessed controls are— but I daresay that we haven't blessed this repo as an official set of controls yet.
3. We need to get more input from folks at diverse institutions, but that part could be done through an iterative review and ongoing refinement of the controls in the repo
4. Beyond the scope of accuracy in the definitions, we will also need to refine the entire repo (likely in an ongoing effort) to adhere to a corporate voice in things such as control descriptions
5. Are we happy merging this in an iterative state, so long as issues are created or updated for the subsequent refinement work?

AdrianHammond:
@eddie-knight  - I am happy to merge on basis we are iterating on this. @mark-rushing FYI

AdrianHammond (approved):
approved and merge - likely to require further iteration but a great start by @smendis-scottlogic

PR #147, opened 2024-03-20
Add taxonomy.md for virtual machines

jisted-scottlogic:


services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+use in financial services ecosystems.
+
+## Taxonomy
+
+| Taxonomy ID | Feature | Description |
+| ----------- | ------- | ----------- |
+|	CCC-030101	|	Instance Types	|	Offers a range of instance types with different specifications for CPU, GPU, memory, storage optimisation and networking capacity.	|
+|	CCC-030102	|	Operating System Options 	|	A selection of operating systems for virtual machine instances.	|
+|	CCC-030103	|	Ephemeral Storage	|	Temporary storage available to the VM which is lost when the instance is stopped or terminated	|
+|	CCC-030104	|	Availability	|	Ensuring high availability of virtual machine instances through redundancy and multiple (availability) zones within a region.	|
```
smendis-scottlogic, line 14: Should the feature be High Availability

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+
+This _service-level taxonomy_ documents the minimal set of features
+that should be present for a service to be considered portable for
+use in financial services ecosystems.
+
+## Taxonomy
+
+| Taxonomy ID | Feature | Description |
+| ----------- | ------- | ----------- |
+|	CCC-030101	|	Instance Types	|	Offers a range of instance types with different specifications for CPU, GPU, memory, storage optimisation and networking capacity.	|
```
smendis-scottlogic, line 11: I have second thoughts how validation tests could be written to test this. Instead should we have like
General Computing
High Performance Computing
Storage Optimized Computing
Low Latency Network Computing
Would like to know what others think of this.

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030102	|	Operating System Options 	|	A selection of operating systems for virtual machine instances.	|
```
smendis-scottlogic, line 12: Same as my comment on CCC-030101

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030103	|	Ephemeral Storage	|	Temporary storage available to the VM which is lost when the instance is stopped or terminated	|
```
smendis-scottlogic, line 13: What about the ability to plug in Permanent Storages. Like EBS with (with delete with instance option turned off) and EFS network drives in AWS.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Makes sense, will change.

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030104	|	Availability	|	Ensuring high availability of virtual machine instances through redundancy and multiple (availability) zones within a region.	|
+|	CCC-030105	|	Identity and Access Management	|	Implementing identity and access management features such as key pairs, JIT and MFA to control user access to virtual machine instances.	|
+|	CCC-030106	|	Monitoring and Logging	|	Offering monitoring and logging capabilities to track performance metrics, user access and security events.	|
+|	CCC-030107	|	Backup and Restore	|	Providing backup and disaster recovery solutions for virtual machine instances and associated data, including snapshot-based backups, incremental backups, and point-in-time recovery.	|
+|	CCC-030108	|	Encryption at Rest	|	Encrypting data stored by virtual machine instances to protect against unauthorized access.	|
```
smendis-scottlogic, line 18: Isn't this a feature on EBS? VM doesn't have a storage of it's own.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Agree that there's potentially a lot to validate by leaving this as one category. A couple of these features are definitely quite hefty, I'll work on splitting this out.

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030109	|	Encryption in Transit	|	Encrypting data transmitted to and from virtual machine instances to protect against interception.	|
```
smendis-scottlogic, line 19: What does VM do to enable encryption on transit?

rgriffiths-scottlogic (commented):
I think you've got some of your Azure and Google references mixed up!

services/compute/virtual-machines/taxonomy.md
rgriffiths-scottlogic: Different providers have different views on what constitutes HPC or HiMem etc.
It might need to be some more fuzzy requirement!

smendis-scottlogic: Should we wait and see what the Taxonomy WG thinks about this?

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030120	|	Instance Health Checks	|	Providing features for performing health checks on virtual machine instances and automatically replacing or repairing unhealthy instances.	|
+|	CCC-030121	|	Instance Remote Access	|	Offering remote access to virtual machine instances through methods such as SSH or RDP for troubleshooting, debugging, and maintenance purposes.	|
+|	CCC-030122	|	Instance Live Migration	|	The ability to perform live migration of virtual machine instances between physical hosts for maintenance or load balancing purposes without downtime.	|
+|	CCC-030123	|	Instance Remote Configuration	|	Providing tools for remotely configuring virtual machine instances, including deployment automation and configuration management frameworks.	|
+|	CCC-030124	|	Instance Resource Tagging	|	Enabling tagging of virtual machine instances with metadata for organization, management, and cost allocation purposes.	|
+|	CCC-030125	|	Instance Resource Monitoring	|	Providing tools for monitoring resource utilization and performance metrics for virtual machine instances, including CPU usage, memory usage, disk I/O, and network traffic.	|
```
smendis-scottlogic, line 35: May be this is just my lack of knowledge on VMs. But how is this different to CCC-030106?

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030126	|	Custom Images	|	Allows users to create and manage their own customized virtual machine images.	|
```
smendis-scottlogic, line 36: Should there be another feature to cover bootstrap scripts that is pre configured to run on start

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030127	|	Dynamic Performance	|	Providing "burstable" instances for intermittent workloads that accumulate credits during periods of low usage which can be used to burst above baseline performance when needed.	|
+|	CCC-030128	|	Dedicated Instances	|	Providing the option to run instances on physical servers that are dedicated solely to a single customer account, ensuring that the underlying hardware resources are not shared with other customers.	|
+|	CCC-030129	|	Instance Scheduling	|	Offering features for scheduling the startup, shutdown, and maintenance of virtual machine instances based on predefined schedules or conditions.	|
+|	CCC-030130	|	Interoperability with Storage Options	|	Capability to read/write to non-ephemeral external storage including object storage and encrypted block storage.	|
+|	CCC-030131	|	Instance Autoscaling	|	Automatically adjusting the number and instance type of virtual machine instances based on predefined criteria such as CPU utilization or incoming traffic,	|
+|	CCC-030132	|	Instance Grouping	|	Offering logical grouping and management tools for sets of virtual machine instances.	|
```
smendis-scottlogic, line 42: This should live above VMs. (May be they use something like tagging to implement this in VM, which is covered in CCC-030124

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030126	|	Custom Images	|	Allows users to create and manage their own customized virtual machine images.	|
+|	CCC-030127	|	Dynamic Performance	|	Providing "burstable" instances for intermittent workloads that accumulate credits during periods of low usage which can be used to burst above baseline performance when needed.	|
+|	CCC-030128	|	Dedicated Instances	|	Providing the option to run instances on physical servers that are dedicated solely to a single customer account, ensuring that the underlying hardware resources are not shared with other customers.	|
+|	CCC-030129	|	Instance Scheduling	|	Offering features for scheduling the startup, shutdown, and maintenance of virtual machine instances based on predefined schedules or conditions.	|
+|	CCC-030130	|	Interoperability with Storage Options	|	Capability to read/write to non-ephemeral external storage including object storage and encrypted block storage.	|
+|	CCC-030131	|	Instance Autoscaling	|	Automatically adjusting the number and instance type of virtual machine instances based on predefined criteria such as CPU utilization or incoming traffic,	|
```
smendis-scottlogic, line 41: Done by Auto Scaling Group in the other hand VM should provide a start/stop API interface for the Auto Scaling Group to work on, which is a feature of the VM.

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030124	|	Instance Resource Tagging	|	Enabling tagging of virtual machine instances with metadata for organization, management, and cost allocation purposes.	|
+|	CCC-030125	|	Instance Resource Monitoring	|	Providing tools for monitoring resource utilization and performance metrics for virtual machine instances, including CPU usage, memory usage, disk I/O, and network traffic.	|
+|	CCC-030126	|	Custom Images	|	Allows users to create and manage their own customized virtual machine images.	|
+|	CCC-030127	|	Dynamic Performance	|	Providing "burstable" instances for intermittent workloads that accumulate credits during periods of low usage which can be used to burst above baseline performance when needed.	|
+|	CCC-030128	|	Dedicated Instances	|	Providing the option to run instances on physical servers that are dedicated solely to a single customer account, ensuring that the underlying hardware resources are not shared with other customers.	|
+|	CCC-030129	|	Instance Scheduling	|	Offering features for scheduling the startup, shutdown, and maintenance of virtual machine instances based on predefined schedules or conditions.	|
```
smendis-scottlogic, line 39: My knowledge is not very sharp on this. Please correct me if I'm wrong. Start/Stop instance base on a schedule should live in a higher level like in a separate Scheduler service. Start/Stop scripts on a instance is part of cron jobs (I assume we can still have a VM which does not support cron jobs and still be able to call it a VM). Maintenance base on schedule should be a feature of a external patching service, etc. (not a feature of VM). Just my thought

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Here I was referring to the availability of hardware metrics, (CPU and memory utilisation % etc) rather than just logging in general but I understand that isn't very clear.
It looks like this is [more of a Cloudwatch feature anyway](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html), but the ec2 would need to provide those metrics so I'm not sure whether to keep this in some form.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Is that covered by **Interoperability with Storage Options** below?

services/compute/virtual-machines/taxonomy.md
smendis-scottlogic: Yes make sense. Should we consider renaming the feature to **Instance Resource Utilization Metrics** or something with Metric instead of Monitoring and change description accordingly?

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: I was going to just delete this CCC-030132 based on this comment but I double checked, and this seems to be one of those where the implementation differs between providers. The grouping does live above EC2 in AWS, but on GCP the Instance Groups are treated as part of Compute Engine itself: https://cloud.google.com/compute/docs/instance-groups.

That confuses things a bit because the GCP Instance Groups also perform some of the features of an aws auto-scaling group or an azure scale set, but maybe for our purposes maybe we could cover the Instance Groups in the same taxonomy as the other two?

jisted-scottlogic:
> I think you've got some of your Azure and Google references mixed up!

@rgriffiths-scottlogic 
I've maybe just spent too long looking at it but I can't see what you mean?

rgriffiths-scottlogic:
> > I think you've got some of your Azure and Google references mixed up!
> 
> @rgriffiths-scottlogic I've maybe just spent too long looking at it but I can't see what you mean?

Line 45 of cloud-services.md and onwards...

Just flip around the two instances of Azure and Google products

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: I've removed the vertical and horizontal scaling as features, and addressed providing an interface for the autoscaling services in a separate point.

jisted-scottlogic:
> > It's not @jisted-scottlogic's code, it's my code. I have mixed that up. @jisted-scottlogic it would be great if you can fix it as part of your PR. Need to flip content of Azure and Google Cloud in the below code which you can find in `cloud-services.md`
> > 
> > * **AWS**: AWS Batch
> > * **Azure**: Cloud Dataflow
> > * **Google Cloud**: Azure Batch
> 
> That explains it!
> 
> There are two distinct instances, I think.

That clears things up, thanks. Switched them round.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Be interested to hear people's thoughts on this. To summarise: if Google treats Instance Groups as 'part of' Compute Engine rather than a separate service, do we need to capture the logical grouping and autoscaling features of Instance Groups as part of the VM taxonomy, or treat them the same as AutoScaling groups/Scale Sets regardless?

services/compute/virtual-machines/taxonomy.md
```diff
@@ -0,0 +1,47 @@
...
+|	CCC-030131	|	Instance Autoscaling	|	Automatically adjusting the number and instance type of virtual machine instances based on predefined criteria such as CPU utilization or incoming traffic,	|
+|	CCC-030132	|	Instance Grouping	|	Offering logical grouping and management tools for sets of virtual machine instances.	|
+|	CCC-030133	|	Scalability - Horizontal	|	Ability to scale virtual machine instances horizontally by adding more instances to accommodate increased workload demands	|
+|	CCC-030134	|	Scalability - Vertical	|	Ability to scale virtual machine instances vertically by adjusting CPU, memory, and storage resources.	|
+|	CCC-030135	|	Security Groups	|	Ability to configure security groups or firewalls to control inbound and outbound traffic to and from instances.	|
+|	CCC-030136	|	Interoperability with Networking Options	|	Offering networking options such as virtual private networks (VPNs), direct connections, and interconnects for connecting virtual machine instances to on-premises networks or other cloud services.	|
```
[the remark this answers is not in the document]
jisted-scottlogic: I've talked about the multiple redundant networking options elsewhere now, and removed this one.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: I was thinking the ephemeral storage can be encrypted? I'll double check though, I imagine it needs integrating with KMS or similar.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Just checking, do you mean you I could list the potential OSs as separate features? If so, how should we determine which ones are required?

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: That's about right for an instance template - you're pre-determining instance type, specs and networking among other things. I've given bootstrap scripts their own point now.

services/compute/virtual-machines/taxonomy.md
rgriffiths-scottlogic: I think that listing out available options is just asking for trouble! Perpetual updates required... for ever! I think stating that multiple OSes *could* or *are* supported (without naming them) should suffice.

services/compute/virtual-machines/taxonomy.md
rgriffiths-scottlogic: Can you term everything in terms of a set of compute? If the set contains >1 entry, it's likely referring to GCE whereas if the set has exactly one entry, then it's a plain old VM from any of the hyperscalers.
That should allow the set to own the scaling properties for n>1 but would also allow for scaling sets to exist separately for n=1.
?????

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: I put this in early on because obviously encryption=essential, but I agree that it's not really the VM doing this part. I think I'll remove this one.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Poked around a bit more to check these and I agree that this scheduling is largely handled by other services. There's an argument to be made with GCP that the instance groups are part of the VM solution (they can be used to establish schedules as well as auto-scaling and grouping), but we have another thread on that issue above for people to weigh in on whether those belong in this taxonomy. For now I'll remove this feature, but I'll back up the markdown somewhere in case we want it back.

AdrianHammond (commented):
Just reviewed and this LGTM.

services/compute/virtual-machines/taxonomy.md
jisted-scottlogic: Still happy to accept a different solution here but I think this question was largely based on a difference in how these features are _named_ , rather than whether or not the feature is available for VMs on each platform.

For now I've left the 'Instance Grouping' and 'Instance Autoscaling' interfaces in here, which should cover these capabilities while remaining platform agnostic - and still allows Autoscalers to be approached separately with less risk of overlap.

damienjburks (commented):
Great work on this @jisted-scottlogic! Exceptionally comprehensive. I think this PR should be approved as a first step. 😄🎊

However, I think we should create an issue revolving around this question for more research to be done:
- Should we create a subset of the most used taxonomy features? 
  - If so, how do we identify which taxonomy features will be included in that subset? We may need to do additional research on this.
  - e.g Users in AWS may not prefer to use General/Storage optimized EC2 instances, whereas GCP users may prefer to use GPU/AI-focused instances or compute engines.

AdrianHammond (approved):
Discussed at 04/04 All hands and merging as discussed in meeting. Opening up new issues as per @damienjburks comment.

Issue #139, opened 2024-03-07
Blocker Clearing for OSCAL & Related Design Questions


eddie-knight:
# Feature Request

This was drafted in collaboration with @iMichaela.

This may relate to #131.

## Description of Problem:

We currently have a blocker for OSCAL development stemming from a lack of clarity around the FINOS' assurance process which needs to be supported by the information captured in the CCC catalog and the way the information is represented in OSCAL.

## Potential Solutions:

We should have a dedicated design session with the maintainers, steering committee, and any interested community members to dive into this issue and clear the blocker.

Before the design session, decisions need to be made and considerations need to be addressed for the design session to be efficient, and for all of us, collectively, to move the project forward.

### Additional Considerations

**Decisions that would trigger different approached for representing the catalog in OSCAL**: 
- Content that needs to be represented:
  - [ ] 1.Threats mitigated by logical controls
  - [ ] 2. Mitigations implemented by logical controls
  - [ ] 3. Logical Controls
  - [ ] 4. Guidance (generic implementation guidance) - Note: service specific guidance can be captured in a Component Definition and not a Catalog.
  - [ ] 5. Examples (implementation examples)
  - [ ] 6. Control objectives (useful in assessing the outcome of a control)
  - [ ] 7. Assessment methods accepted for each control
  - [ ] 8. Something else (e.g. metrics)
- Assessment process details that can affect the way the information is represented
  - [ ] A certification/assessment process will only focus on the logical controls and report their status 
    - This approach can simplify the CCC Catalog by focusing only on the logical controls and related info (3 and after from the above list) and a resilience analysis can be performed (automatically) using either a "Catalog" of Mitigations annotated with the Threats info and links to MITRE ATT&CK, mapped in OSCAL to the CCC catalog OR Threats & Mitigations - mapped to - CCC Controls.  The distinction is in how the reporting is expected (e.g. which mitigations are properly implemented OR which threats are properly defended)
  - [ ] A certification/assessment process will extend to analyzing the resilience of the system incorporating the threats and mitigation.
    - This approach will have to have all the necessary information in one OSCAL CCC Catalog (all 1 through 8, from the above list).
- Other OSCAL information: Component Definitions for each service. (The following decision needs to be made) FINOS intends to generate and maintain:
  - [ ] OSCAL Profiles for each service 
    - (aka the list of CCC Controls a service needs to implement and any necessary tailoring of the controls for each service). 
[...]

Issue #126, opened 2024-02-13
02/15/2024 Common Cloud Controls - MITRE WG


crawfordchanel:
## Date
02/15/2024 - 10 AM ET / 3 PM UK

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies). 

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda
- [x] Convene & roll call (5mins)
- [x] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) 
- [x] Review Meeting Notices (see above)
- [x] Approve past meeting minutes
- [ ] #11 
- [ ] #14 
- [ ] #75 
- [ ] #77 
- [ ] AOB, Q&A & Adjourn (5mins)

### Zoom info

Join Zoom Meeting
https://zoom.us/j/93861901920

Meeting ID: 938 6190 1920
Passcode: 284383

---

Dial by your location
• +1 719 359 4580 US
• +1 253 205 0468 US
• +1 253 215 8782 US (Tacoma)
• +1 301 715 8592 US (Washington DC)
• +1 305 224 1968 US
• +1 309 205 3325 US
• +1 312 626 6799 US (Chicago)
• +1 346 248 7799 US (Houston)
• +1 360 209 5623 US
• +1 386 347 5053 US
• +1 507 473 4847 US
• +1 564 217 2000 US
[...]

Issue #11, opened 2023-08-31
Contribution Request : OSCAL example that points to MITRE and describes tests using Gherkin


mcleo-d:
## Description
During #5 the following requests were raised by the CCC MITRE working group.

- [ ] Can Citi contribute an existing OSCAL example that can be used as a reference by the MITRE working group?
  - This example could be used to inform further MITRE contributions through OSCAL.
  - Can this contribution be raised as a PR into the CCC repo?

- [ ] The work related to `Storage ( object storage ) [AWS S3, gcp storage, azure object storage]` is complete and ready to be contributed into the FINOS CCC repo 
  - https://github.com/finos/common-cloud-controls/issues/11#issuecomment-1729746032

--- Update ---

The following item has been moved into its own separate issue as it creates a larger scope.
- Gherkin should be used as a `Behaviour Test` rather than a `Configuration Test`. 
  - Can an example of this type of test be expanded upon during a relevant future call?

AdrianHammond:
@eddie-knight  - this look good. 

@mark-rushing @moematar @simonzhangbmo @vicenteherrera  - what do you think? Are you okay with this and should we #72 ?

mark-rushing:
It looks great! Think it would be good to run services/database/relational/rdms-taxonomy.feature past the other two working groups <- @jonmuk @git-hub-forwork1


From: "[github.com] Adrian Hammond" ***@***.***>
Reply-To: finos/common-cloud-controls ***@***.***>
Date: Tuesday, October 24, 2023 at 2:23 AM
To: finos/common-cloud-controls ***@***.***>
Cc: "Rushing, Mark [CCC-OT]" ***@***.***>, Mention ***@***.***>
Subject: Re: [finos/common-cloud-controls] Contribution Request : OSCAL example that points to MITRE and describes tests using Gherkin (Issue #11)

@eddie-knight - this look good. @mark-rushing @moematar @simonzhangbmo @vicenteherrera - what do you think? Are you okay with this and should we #72 ? — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving


@eddie-knight<https://urldefense.com/v3/__https:/github.com/eddie-knight__;!!Jkho33Y!iLEEHhLyqR_NxpRdRQZqG4PfuW2yDBRQfQpuJor3uCsotc-6jktUzgJQnpZH5t5M6qBG-IXZDkRnY1GvZccRQgZG$> - this look good.

[...]

PR #153, opened 2024-04-04
Draft Proposal for Threat Catalog and Control Catalog Taxonomy 

mlysaght2017:
This PR is draft and is an initial proposal from Damien and myself on structure to a basic threat catalog and control catalog taxonomy in basic md format, with example mappings to the generic CCC object storage service as a starting point

The PR addresses some of the questions posed in #139 as well as #11 within #126


mlysaght2017:
@eddie-knight @rowan-baker @jonmuk @AdrianHammond @iMichaela  - can you please add yourself as reviewers so we can keep the momentum on feedback. Please add anyone else that makes sense.

The overview  of the PR at the last all-hands call was well received with good positive feedback, but would like to keep momentum on the PR review so that we can start iterating on the respective threat and control catalogs and keep overall momentum in parallel to more formalisation on the WGs via the steerco

Thanks!

eddie-knight:
This may need to be captured elsewhere instead of blocking the PR...

@robmoffat and I were brainstorming previously regarding the scope of work and the need to reuse existing material. With the values proposed here, how much do we anticipate we will be able to reuse tools from folks such as [MITRE Engenuity](https://github.com/center-for-threat-informed-defense)? 

Relatedly, is there a way we can confirm that our structure is best suited to harness work that's already been done by the greater cybersecurity community?

AdrianHammond (commented):
Hi @mlysaght2017 thanks for picking this up and driving 

I like the idea of having top level info to help users navigate into CCC as I see the most common user entry point is going to be the cloud service taxonomy that is being consumed.  

I just have a couple of comments / questions: 

1. Do we need all of this information here (objective, description and test) as this could become a maintenance overhead in the future?  I think having just the Objective would be sufficient. 
2. Is Control id and Service Taxonomy always going to have a 1-2-1 relationship? If so should we use Service Taxonomy id as top level id?
3. Should we take this opportunity to update the format of the Service taxonomy id to CCC-xx-nnnn where xx is 2 characters representing the service, like you have used in control id

control-catalog/storage/object/controls.md
damienjburks, line 1: @mlysaght2017 do you think it'd be a better idea to move the `controls.md` & `threats.md` files into the `services/storage/object` folder? I think it'll make things cleaner overall, and we can easily map the Service Taxonomy IDs to reference the taxonomy.md file. 

For example:
- services/storage/object
   - taxonomy.md
   - controls.md
   - threats.md

mlysaght2017: I like this idea @damienjburks - as you say makes things cleaner and easier to manage

mlysaght2017: I've moved the threat and control md files into the services folder

control-catalog/storage/object/controls.md
```diff
@@ -0,0 +1,5 @@
+| Control Id | Objective | Description | Test | Service Taxonomy Id | NIST CF | MITRE ATT&CK Mitigations | Threats |
```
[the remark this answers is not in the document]
mlysaght2017: This is an interesting question. On `Detect`, I can see benefit in assessing the CSP's native threat detection capabilities (as well as basic audit logging of specific APIs) for a given service - e.g. data stores exposed outside of given data perimeters - here we'd also have the option of mapping to MITRE ATT&CK `Data Sources` e.g. for Cloud Storage: https://attack.mitre.org/datasources/DS0010/  

However...this would naturally be a lot more work...It feels like our priority should be on maturing our thinking on `Protect` with more `Protect` examples within and across the services and continuing to think/experiment on `Detect`? Feels like an early decision for steerco too.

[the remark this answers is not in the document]
mlysaght2017: Totally agree that the `Test ` values in Gherkin don't read well in that format. We can shift to a separate gherkin file as you suggest.

mlysaght2017:
Thanks @AdrianHammond 

1. Do we need all of this information here (objective, description and test) as this could become a maintenance overhead in the future? I think having just the Objective would be sufficient.

ML: 

- We're discussing shifting the `Test` out into a separate gherkin file - see discussion with @eddie-knight above.
- I think we can also remove the `Service Taxonomy Id` as this is already contained within the separate threats.md file (i.e. the threat catalog) - in effect a duplication 
- `Objective`/`Description`: I'm pretty sure I agree with you here In my head right now, individual logical controls underly a high level control `Objective` with a 1-to-many relationship between control `Objective` and controls `Description` in effect. Maybe we can stick to just `Objective` and see how it goes as we think more about the "Assessment Layer"
 
3. Is Control id and Service Taxonomy always going to have a 1-2-1 relationship? If so should we use Service Taxonomy id as top level id?

ML: I can see the Service Taxonomy having a 1-2-Many relationship with the Threats


5. Should we take this opportunity to update the format of the Service taxonomy id to CCC-xx-nnnn where xx is 2 characters representing the service, like you have used in control id

ML: I think there's been a new issue created to standardize on this - will check, but either will align asap.

mlysaght2017:
Thanks @eddie-knight - all great points!

> This may need to be captured elsewhere instead of blocking the PR...
> 
> @robmoffat and I were brainstorming previously regarding the scope of work and the need to reuse existing material. With the values proposed here, how much do we anticipate we will be able to reuse tools from folks such as [MITRE Engenuity](https://github.com/center-for-threat-informed-defense)?
> 
> Relatedly, is there a way we can confirm that our structure is best suited to harness work that's already been done by the greater cybersecurity community?

Realise we've had a couple of discussions on the mappings MITRE Engenuity have done between ATT&CK and NIST 800-53 Rev 4...
Thinking here is that while the project scope continues to get ironed out we continue to provide examples that will help with some of the decision making. It still feels like we need more examples of what <Service> -> <Threats> -> <Controls> -> <Assessment> would look like and at the same time continue to investigate what's out there already that we can leverage (both in terms of  catalogs and mappings)...?

services/storage/object/threats.md
iMichaela, line 1: The `threats.md` file is mixing locally defined threats with MITRE ATT&CK ones and the inconsistency will trigger digitalization (OSCAL representation) difficulties or barriers.

iMichaela (approved):
I see no comments or changes to OSCAL representation options.

eddie-knight (approved):
Adding my approval, since all of the salient comments have been addressed.

@mlysaght2017 — Understanding that this is not intended for delivery, would you like to merge this as an iterative step toward the next deliverable?

sshiells-scottlogic (approved):

damienjburks (approved):

PR #179, opened 2024-05-20
Add a warning on changing FINOS hosted meetings

sshiells-scottlogic:
Add a  note to remind people that if adjusting/cancelling a FINOS hosted meeting that people should include a FINOS employee when making the changes.

### DIFF
--- a/docs/governance/community-guidelines/communication.md
+++ b/docs/governance/community-guidelines/communication.md
@@ -36,7 +36,7 @@ Any meeting published on the public calendar must additionally adhere to a stric
 - If these meetings are hosted by FINOS they must follow the guidance for [FINOS hosted meetings](#finos-hosted-meetings).
 - If these meetings are NOT hosted by FINOS then any noteworthy decisions or outcomes should be communicated back to the wider [WG] via the mailing list.
 
-[SC]: <../../community-groups.md#steering-committee>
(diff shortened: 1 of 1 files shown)


sshiells-scottlogic:
@eddie-knight can you review? I am not sure if FINOS employee is the correct term. I think we used "point of contact" in our discussion?

eddie-knight:
You're right @sshiells-scottlogic— the phrasing should be "point of contact." The mistake last week was that we only went through FINOS employees, unreasonably placing the responsibility on them to coordinate internally when we could/should have coordinated with our POC prior to making the request.

cc @robmoffat

sshiells-scottlogic:
I will update. Should we keep it generic, or list Rob specifically?

eddie-knight:
Good shout. We haven't decided on where to tract the FINOS POC in the repo.

For now let's link to the role definition: https://github.com/finos/common-cloud-controls/blob/main/docs/governance/steering/charter.md#finos-point-of-contact

As a follow-up point... We should also maintain a list of who fills which leadership roles in the README or elsewhere (governance/README?).

sshiells-scottlogic:
I have updated this PR with the changes as described (and fixed a couple of broken links). I will have a think about who fills what roles and put a potential solution out when ready.

PR #191, opened 2024-06-06
Fix link to delivery wg in taxonomy charter

sshiells-scottlogic:
Fix the link to the delivery working group in the taxonomy working groups charter.


eddie-knight (approved):

eddie-knight:
Nice to see the codeowners adding the WG! But... in this case, I think the intent was to require approval from the SteerCo to modify charters.

🤔 https://github.com/finos/common-cloud-controls/blob/main/.github/CODEOWNERS#L26

Good to merge when it lets you, but we should fix this edge case.

PR #244, opened 2024-07-08
Added primer

robmoffat:


eddie-knight (changes requested):
This looks really cool, @robmoffat. I'd like to also get a review from the Comms WG to confirm we're all on the same page as it goes into the repo.

eddie-knight (approved):

eddie-knight:
Got a thumbs up from @Alexstpierrework. Thanks for putting this together @robmoffat!

PR #269, opened 2024-07-22
fixed primer path

robmoffat:


eddie-knight:
Hey @robmoffat, could we move this to the `docs/` repo?

I'm not sure `formation` is a good fit 🤔 Maybe we need to adjust the strucutre... but somewhere in docs seems appropriate. WDYT?

robmoffat:
How about this?  Hard to say if the link will work until we've pushed it

sshiells-scottlogic (approved):
Happy to approve/merge and we can fix the link if is broken?

PR #279, opened 2024-07-25
Update CODEOWNERS for docs/resources

eddie-knight:


eddie-knight:
@robmoffat two issues here that seem like non-issues:

1. The new lint/fmt checks appear to be overzealous at the moment, and don't apply to this PR.
2. I'm pretty positive that the non-existant group error presented in the review UI is incorrect

jared-lambert (approved):

PR #280, opened 2024-07-25
Update readme.md to mention CDMC in the controls section

ojeb2:
simple description of CDMC. plan to add a better one in the near future which will include links to docs & training material


jared-lambert (approved):
LGTM

PR #281, opened 2024-07-26
Minor fixes to the CODEOWNERS file

eddie-knight:
I'm still not sure that our teams have the correct permissions, but I'm hoping this makes the directory ownership work.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,8 +6,7 @@
 #
 # NOTE: If the GitHub automated reviewers do not behave as expected,
 #       please contact the Community Structure WG.
-* @finos/ccc-wg-community-structure
-.github/CODEOWNERS @finos/ccc-wg-community-structure
+/* @finos/ccc-wg-community-structure
 #
 ########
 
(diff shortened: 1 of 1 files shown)


eddie-knight:
The check failures don't apply to changes in this PR

eddie-knight:
@sshiells-scottlogic @damienjburks Can either of yall take a look at this?

I like that merging is a lot faster with CODEOWNERS broken 😆 but I'd prefer we refine the codeowners usage instead of just leaving it broken. Hoping this gets it turned back on, then we can refine it more later.

damienjburks (commented):
@eddie-knight this is the exception that I'm getting:

```
Unknown owner on line 17: make sure the team @finos/ccc-wg-delivery exists, is publicly visible, and has write access to the repository
```

it seems to be repeated for every working group... I'm assuming the permissions for each team is already created properly right?

eddie-knight:
@damienjburks I suspect there is an issue with the teams settings, and this PR may be a red herring, but I'd like to rule this out as a contributing factor before we set up time with FINOS to investigate all of the teams.

sshiells-scottlogic:
@eddie-knight  are you just looking for an approval here? Or to look into the actual issue?

eddie-knight:
@sshiells-scottlogic Looking to merge this and confirm whether it impacts the issue. Then I'll swing back around with help@finos to figure out the teams settings.

damienjburks (approved):

damienjburks:
@eddie-knight okay - merging this in now then.

PR #284, opened 2024-07-29
Review and update WG meeting issue templates

dogle-scottlogic:
- Change headers to match WG instead of "All Hands"
- Add regular times to save WG leads having to fill out - use language to indicate that these are the standard times


sshiells-scottlogic (approved):

eddie-knight (approved):
Thanks for making this fix!

Issue #299, opened 2024-08-05
Adjust CCC development format to create automation opportunities


eddie-knight:
Initial contributions have highlighted the complexity of maintaining Markdown documents.

## Background

- This project intends to release consistently formatted outputs
- An iterative decision was made to create all documents in Markdown format
- Contributions such as Component Definitions have a high degree of complexity
- Minor inconsistencies have arisen in the markdown formatting

## Problem

There is currently no method for automatically parsing CCC content into OSCAL or CFI-compatible assets.

## Proposed Solution

Contributors will content in YAML first, then at release time these will be programmatically parsed into Markdown and OSCAL assets.

- **Why YAML?** — Because it is a simplified machine readable format, designed to also be "human readable"
- **What assets will be written in YAML?** — The current discussion is focused on Component Definitions, but we may consider extending the approach to Taxonomy assets if it is found to create value in that area as well.

PR #271, opened 2024-07-24
Added guidelines and template for control definitions.

eddie-knight:
Attempt at codifying the discussion in the comments of #263

PR #300, opened 2024-08-05
Added yaml to md generator for component definitions

eddie-knight:
Part of #299.

This go script can be used to generate a markdown file from a YAML asset that matches a particular format. 

command example:

`go run yaml-md-generator.go ../services/storage/object/controls.yaml`

#271 documents the expected yaml structure.

Example input and output are attached. (yaml is in txt format because of attachment restrictions)

[controls.yaml.txt](https://github.com/user-attachments/files/16499582/controls.yaml.txt)
[CCC.OS.md](https://github.com/user-attachments/files/16499592/CCC.OS.md)


damienjburks:
@eddie-knight it might be a good idea to add the `go.sum` file into the .gitignore and remove it from this repo. Thoughts?

eddie-knight:
> @eddie-knight it might be a good idea to add the `go.sum` file into the .gitignore and remove it from this repo. Thoughts?

I did a crazy check on reddit and stack overflow, and it seems to me as if the general consensus is to track the `go.sum` in the repo. Also... it's kindof like a dollar-store SBOM! So not completely without value.

AdrianHammond (commented):
Should we name the `release` directory `release tooling` ?

PR #326, opened 2024-09-02
Add in proof-of-concept oscal profile for object storage

mlysaght2017:
As part of proof-concept, this YAML structure organizes controls into an _OSCAL Profile_ in the OSCAL _Control Layer_  with respective groups based on their control families, and each control includes an assessment-objective section that maps directly to testing requirements.

### DIFF
--- a/services/storage/object/oscal-profile.yaml
+++ b/services/storage/object/oscal-profile.yaml
@@ -1,7 +1,7 @@
 catalog:
-  uuid: "ccc-object-storage-catalog-uuid"
+  uuid: "ccc-catalog-uuid"
   metadata:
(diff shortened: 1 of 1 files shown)


mlysaght2017:
Had working session with @eddie-knight and @abikhuil and came up with example/poc OSCAL profile as a potential target state for representing CCC controls within the _OSCAL control layer_ 

Note that all controls in the profile here are "common" controls - i.e. non service specific. The OSCAL profile example here reflects the resolving of a subset of the a full set of common controls into a profile of controls relevant to the object storage service.

eddie-knight:
This first iteration is still needed as a first step on the path to automate the OSCAL output in our releases.

Issue #364, opened 2024-09-12
Formalise yaml schemas


sshiells-scottlogic:
### Description of Problem:

Currently we have example template files for features, controls, threads at meta data. Having templates instead of schemas make it more difficult to version control and validate.

### Potential Solutions:

Produce yaml schemas based on the current template files.

dogle-scottlogic:
https://github.com/finos/common-cloud-controls/pull/369

iMichaela:
Does it make sense to use the OSCAL YAML schema for validation. In the examples I created long ago, I included a Makefile that invokes the oscal-cli tool for validation and conversion into all supported formats (XML, JSON, YAML). The Makefile can be used for local development to validate oscal artifacts and can be invoked from CI/CD pipeline.

PR #369, opened 2024-09-13
Add yaml schemas (Issue #364)

dogle-scottlogic:
Add YAML schemas for controls, features, metadata and threats. Tested locally with VSCode

Using draft-07 as it seems most widely supported

image

### DIFF
--- a/docs/governance/community-guidelines/templates/schemas/README.md
+++ b/docs/governance/community-guidelines/templates/schemas/README.md
@@ -1,0 +1,16 @@
+# Yaml Validation Schemas
+Yaml validation schemas have been created for controls, features, metadata and threats files. 
+
+## VSCode integration
+You can update VSCode to highlight issues using the schema files with the following steps:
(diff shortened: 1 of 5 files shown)


sshiells-scottlogic (dismissed):
Looks great Dave. Would it be worth adding some instructions on how to use the schemas within VS code until we get a more automated way?

dogle-scottlogic:
can do - where is best to add that detail?

dogle-scottlogic:
There are some fails when I hook up the validator to this. Not sure if we want to update the template, update the failing yml or else allow additional fields in the schema.
https://github.com/finos/common-cloud-controls/pull/372
Might be one to fix in the follow-up PR.

eddie-knight:
We're getting nested really deep in the repo, but it's not an issue for now.

@sshiells-scottlogic, let's swing through with a follow-up action later to put a closer eye on how we want to organize this.

eddie-knight (approved):
LGTM— if there are any issues we can sort those out in a follow-up PR

PR #370, opened 2024-09-13
Add initial gitvote config

sshiells-scottlogic:
Add intial gitvote config with some best guess parameters for what we want:

Vote duration: 5 days
Periodic automatic status check: 1 day
Pass Threshold: 50% (i.e. majority)
Vote closes on passing: true
Binding votes: SteerCo

### DIFF
--- a/docs/.gitvote.yml
+++ b/docs/.gitvote.yml
@@ -1,0 +1,15 @@
+profiles:
+  default:
+    # Voting duration of 5 days
+    duration: 5d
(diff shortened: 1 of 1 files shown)


sshiells-scottlogic:
I am not sure if we can test before this is merged, but once merged we should be able to test on an issue as this should allow votes on both PRs and Issues.

eddie-knight (approved):

PR #373, opened 2024-09-13
Update to common features and service families

smendis-scottlogic:
This PR will contain necessary base files for the KMS work to build upon.

### DIFF
--- a/services/common-features.yaml
+++ b/services/common-features.yaml
@@ -32,3 +32,52 @@ features:
     description: |
       Publishes events for creation, deletion, and modification of
       objects in a way that enables users to trigger actions in response.
+  - id: CCC.F08
+    title: Multi zone deployability
+    description: |
+      Proving the ability for the service to be deployed in multiple availability
+      zones within a region to increase availability and fault tolerance.
+  - id: CCC.F09
+    title: Monitoring
+    description: |
+      Providing the ability to continuously observe, track, and analyze
+      the performance, availability, and health of the service resources or applications.
+  - id: CCC.F10
+    title: Logging
+    description: |
+      Providing the ability to transmit system events, application activities, or/and
+      user interactions to a logging service
+  - id: CCC.F11
+    title: Backup
+    description: |
+      Providing the ability to create copies of associated data or configurations in form of automated backups,
+      snapshot-based backups, or/and incremental backups.
+  - id: CCC.F12
+    title: Recovery
+    description: |
+      Providing the ability to restore data, system, or application to a functional state
+      after an incident such as data loss, corruption or a disaster.
+  - id: CCC.F13
+    title: Infrastructure as Code
+    description: |
+      Providing the automated deployment of the service resources through templates.
+  - id: CCC.F14
+    title: API Access
+    description: |
+      Allowing users to interact programmatically with the sevice and its resources using APIs, SDKs and CLI.
+  - id: CCC.F15
+    title: Cost Management
+    description: |
+      Providing the ability to filter spending and to detect cost anomalies by ther service.
+  - id: CCC.F16
+    title: Budgeting
+    description: |
+      Providing the ability to set a maximum spend for the service.
+  - id: CCC.F17
+    title: Alerting
+    description: |
+      Providing the ability to set an alarm based on performance matrix, logs, events or spending thresholds of the service.
+  - id: CCC.F18
+    title: Versioning
+    description: |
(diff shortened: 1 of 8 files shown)


services/common-features.yaml
```diff
@@ -32,3 +32,52 @@ features:
     description: |
       Publishes events for creation, deletion, and modification of
       objects in a way that enables users to trigger actions in response.
+  - id: CCC.F08
+    title: Multi-zone Deployability
+    description: |
+      Proving the ability for the service to be deployed in multiple availability
```
sshiells-scottlogic, line 38: "Providing" instead of "Proving"?

services/crypto/key/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+title: Key Management
+id: CCC.Key
```
sshiells-scottlogic, line 2: Should we change to something like "CCC.KeyMgmt"? Could jut be me, but just using "Key" doesn't seem very clear to me (same with the containing folder). But I am not a regular visitor to this space, so it may make more sense to those who are more familiar with this area?

services/crypto/service-families.yaml
```diff
@@ -0,0 +1,46 @@
+title: CCC Cryptographic Services
+id: CCC.Crypto
+description: |
+  In cloud computing, cryptographic services refer to cloud-based tools that protect
+  data and communications through encryption and decryption. They also provide secure 
+  methods for storing and managing keys, secrets, and certificates. These services 
+  ensure the confidentiality, integrity, and authenticity of data in both storage and transmission.
+service_categories:
+  - id: CCC.Key
```
sshiells-scottlogic, line 9: If you make changes based on my comment above you would also have to change here.

services/crypto/key/metadata.yaml
rgriffiths-scottlogic: I agree with Stevie here. Making it clear we're talking about Key Management would be useful.

services/common-features.yaml
```diff
@@ -32,3 +32,52 @@ features:
...
+      the performance, availability, and health of the service resources or applications.
+  - id: CCC.F10
+    title: Logging
+    description: |
+      Providing the ability to transmit system events, application activities, and/or
+      user interactions to a logging service
+  - id: CCC.F11
+    title: Backup
+    description: |
+      Providing the ability to create copies of associated data or configurations in form of automated backups,
```
rgriffiths-scottlogic, line 53: in the form
```suggestion
      Providing the ability to create copies of associated data or configurations in the form of automated backups,
```

services/common-features.yaml
```diff
@@ -32,3 +32,52 @@ features:
...
+      snapshot-based backups, and/or incremental backups.
+  - id: CCC.F12
+    title: Recovery
+    description: |
+      Providing the ability to restore data, system, or application to a functional state
```
rgriffiths-scottlogic, line 58:
```suggestion
      Providing the ability to restore data, a system, or an application to a functional state
```

services/common-features.yaml
```diff
@@ -32,3 +32,52 @@ features:
...
+      after an incident such as data loss, corruption or a disaster.
+  - id: CCC.F13
+    title: Infrastructure as Code
+    description: |
+      Providing the automated deployment of the service resources through templates.
```
rgriffiths-scottlogic, line 63: Does IaC only do automated deployment? What about the ability to maintain the intended state? Does that belong here? Or is that better dealt with someplace else?

services/common-features.yaml
```diff
@@ -32,3 +32,52 @@ features:
...
+    description: |
+      Allowing users to interact programmatically with the service and its resources using APIs, SDKs and CLI.
+  - id: CCC.F15
+    title: Cost Management
+    description: |
+      Providing the ability to filter spending and to detect cost anomalies by the service.
+  - id: CCC.F16
+    title: Budgeting
+    description: |
+      Providing the ability to set a maximum spend for the service.
```
rgriffiths-scottlogic, line 75: Most of the budget tools have some kind of threshold levels based on expected spend. Do all of them have a "you can't spend more than this" type of thing? I'm not sure!

sshiells-scottlogic (approved):

smendis-scottlogic:
@eddie-knight @rgriffiths-scottlogic  Need a review from steerco to merge this in

Issue #323, opened 2024-08-30
New service taxonomy definition for Cryptographic Key Management Service (KMS)


mlysaght2017:
### Description of Problem:

We are confident that one of the dependent services for the AI RAG reference architecture will be a cryptographic key management service (Azure Key Vault has so far been explicitly mentioned in #314 ) with data-at-rest encryption expected to be the primary use case for the KMS service within the RAG solution. 

We therefore need a common service definition for a key management service (KMS) so that we can start defining a set of associated common cloud controls. 

### Potential Solutions:
A service taxonomy set of features defined for KMS (covering the common features of AWS KMS, GCP KMS and Azure Key Vault)

PR #384, opened 2024-09-18
Add Key Management Service to CCC service taxonomy (#323)

mic-smith:
Add initial `features.yaml` for the Key Management Service (#323)

### DIFF
--- a/services/crypto/key/features.yaml
+++ b/services/crypto/key/features.yaml
@@ -82,7 +82,7 @@ features:
     title: Soft Delete
     description: |
       Supports the ability to prevent the immediate deletion of a managed key. This includes the ability
-      recover accidental deletion of keys within the grace period.
+      recover accidental deletion of keys within a grace period.
   - id: CCC.KeyMgmt.F19
     title: Delete Key
     description: |
--- a/services/crypto/key/features.yaml
+++ b/services/crypto/key/features.yaml
@@ -14,7 +14,6 @@ features:
     title: AES-256
     description: |
       Support for the AES-256 Advanced Encryption Standard with a 256-bit key for encryption and decryption.
-      Note: Azure only supports symmetric keys in the managed HSM service.
   - id: CCC.KeyMgmt.F02
     title: RSA-2048
     description: |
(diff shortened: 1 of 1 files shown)


services/crypto/key/features.yaml
```diff
@@ -0,0 +1,103 @@
...
+  - CCC.F09 # Monitoring
+  - CCC.F10 # Logging
+  - CCC.F13 # Infrastructure as Code
+  - CCC.F14 # API Access
+features:
+  - id: CCC.KeyMgmt.F01
+    title: AES-256
+    description: |
+      Support for the AES-256 Advanced Encryption Standard with a 256-bit key for encryption and decryption.
+      Note: Azure only supports symmetric keys in the managed HSM service.
```
smendis-scottlogic, line 17: I think note is bit misleading here. For me it will make more sense if it was written as 
Note: AWS and Google Cloud support this algorithm by default, whereas Azure's support for this is limited to its managed HSM (Hardware Security Module) service.

smendis-scottlogic (changes requested):

smendis-scottlogic:
P-521 under ECDSA, AES-128, AES-256, SHA-256, SHA-384, SHA-512 can you check whether these keys are supported by the Key Management services in all 3 cloud service providers as well?

services/crypto/key/features.yaml
eddie-knight: We probably don't want to make point-in-time assertions about CSP features, right?

mic-smith:
> P-521 under ECDSA, AES-128, AES-256, SHA-256, SHA-384, SHA-512 can you check whether these keys are supported by the Key Management services in all 3 cloud service providers as well?

- P-521 -> Not GCP
- AES-128 and AES-192 -> Azure only
- SHA-256, SHA-384, SHA-512. These are just covered implicitly under the RSA encrypt and sign entries at the minute.

The common signing algorithms (with varying key lengths) are:
 - RSASSA_PSS_SHA_256
 - RSASSA_PSS_SHA_512
 - RSASSA_PKCS1_V1_5_SHA_256
 - RSASSA_PKCS1_V1_5_SHA_512

services/crypto/key/features.yaml
```diff
@@ -0,0 +1,101 @@
...
+      Supports the ability to disable a managed key without deletion.
+  - id: CCC.KeyMgmt.F17
+    title: Enable key
+    description: |
+      Supports the ability to re-enable a disabled managed key.
+  - id: CCC.KeyMgmt.F18
+    title: Soft Delete
+    description: |
+      Supports the ability to prevent the immediate deletion of a managed key. This includes the ability
+      recover accidental deletion of keys within the grace period.
```
sshiells-scottlogic, line 85:
```suggestion
      recover accidental deletion of keys within a grace period.
```

sshiells-scottlogic: Just wondering if "the" grace period could be read as a standard grace period, whereas "a" grace period just signals that there is a grace period?

smendis-scottlogic (approved):

PR #388, opened 2024-09-19
Add in updates for traffic light protocol testing requirements

mlysaght2017:


eddie-knight:
cc/ @nas-hub @sublimino @kazmik23

eddie-knight (approved):
Discussed on the Security WG call. If we find opportunities to de-duplicate test requirements, we may adjust this in the near future.

PR #389, opened 2024-09-19
Placeholder for GenAI taxonomy created

smendis-scottlogic:


eddie-knight:
This has a lot of extra changes beyond the scope of the GenAI taxonomy... Would it be too much trouble to update this PR to not include additional changes?

PR #390, opened 2024-09-19
Added example refarch entry

eddie-knight:


mlysaght2017 (approved):
@eddie-knight  - Looks like a good starting point and enough to get going with AI Readiness.

No doubt you're clued into CALM already: https://github.com/finos/architecture-as-code/tree/main - might become more relevant at some point.

eddie-knight:
I did indeed forget to consider CALM as an option. I agree that's the direction we should take this, rather than the ad hoc solution currently in this PR.

PR #406, opened 2024-09-22
Add in new KeyMgmt threat and tests

mlysaght2017:
@eddie-knight - added in todos as well on two threats related to KMS that are currently down as common, but may be better shifted to KMS specific threats - let me know your thoughts.

### DIFF
--- a/services/common-threats.yaml
+++ b/services/common-threats.yaml
@@ -10,7 +10,7 @@ threats:
       - TA0005
       - T1562
   - id: CCC.TH02
-    title: Vendor-hosted keys are compromised
+    title: Vendor-hosted keys are compromised #TODO: Should this be in the crypto service?
     description: |
       The service uses a vendor-hosted key management service (KMS) to manage
       encryption keys. Insider threats or mistakes can result in access by a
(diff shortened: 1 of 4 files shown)


eddie-knight (approved):

damienjburks (dismissed):
LGTM! 🚀

damienjburks (approved):

damienjburks:
@smendis-scottlogic can you please review?

services/crypto/key/threats.yaml
```diff
@@ -1,18 +1,32 @@
 common_threats:
   - CCC.TH01
+  - CCC.TH03
+  - CCC.TH13
+
```
smendis-scottlogic, line 5: We need CCC.TH05 here as CCC.C06 is added under common_controls for KMS and CCC.C06 refers to CCC.TH05

smendis-scottlogic: Add all the referred threats need their title as a comment in front.
Eg - CCC.TH01 # Unauthorized access through elevated privileges

services/common-threats.yaml
```diff
@@ -32,7 +32,7 @@ threats:
       - TA009
       - T1557
   - id: CCC.TH04
-    title: Attacker encrypts data with client-managed keys
+    title: Attacker encrypts data with client-managed keys #TODO: Should this be in the crypto service?
     description: |
       The service provides encryption mechanisms, but the encryption keys are
```
smendis-scottlogic, line 37: `common_threats.yaml` line 54 has a TODO. I think Multi-region deployment make sense to add there. Better to fix it with this MR as CCC.TH05 (line 48-57) should be included in KMS threats.yaml because it is referred by CCC.C06 which is added in KMS controls.yaml

PR #418, opened 2024-09-24
Refactoring common entries

eddie-knight:
- reorganized threats and assigned new IDs (destructive, but created some more intuitive sorting for these early additions)
- added title comments to ID lines, for easier readability while working (ie, folded sections in VSCode) and easier copying to other files later.
- corrected line lengths
- added test requirements
- started adding control mappings, but paused on that for now 

follows #420

PR #419, opened 2024-09-24
Object storage final polish

eddie-knight:
Hopefully this wraps up Object Storage to make it ready for a release candidate.

Follows #418 to reduce the diff


services/common-controls.yaml
```diff
@@ -117,14 +165,34 @@ controls:
...
+        text: |
+          The service logs all access attempts from untrusted entities,
+          including failed connection attempts.
+        tlp_levels:
+          - tlp_amber
+          - tlp_red
+      - id: CCC.C05.TR04
+        text: |
+          The service prevents unauthorized cross-tenant access, ensuring that
+          only allowlisted services from other tenants can access resources.
```
mlysaght2017, line 188: what do we mean by "tenant" here? I think that's worth expanding on.

mlysaght2017 (approved):
Great polish @eddie-knight - LGTM! (one comment on the meaning of "tenant")

damienjburks (approved):

services/common-controls.yaml
eddie-knight: This change isn't in the current PR after merging 418, but I think it's worth spinning back around to if you'd like to propose a change to common controls

services/storage/object/controls.yaml
```diff
@@ -10,14 +10,13 @@ controls:
...
-      Prevent any requests to object storage buckets or objects using untrusted
-      KMS keys to protect against unauthorized data encryption that can impact
-      data availability and integrity.
+      Prevent any requests to object storage buckets or objects using 
+      untrusted KMS keys to protect against unauthorized data encryption 
+      that can impact data availability and integrity.
     control_family: Data
     threats:
       - CCC.TH01
       - CCC.TH02
```
sshiells-scottlogic, line 19: Should we add comments here like we do with common features etc?

sshiells-scottlogic (approved):

services/storage/object/controls.yaml
eddie-knight: Yeah, that's a good idea. I'll aim for another PR with that, but it shouldn't be a blocker for cutting the release candidate.

PR #420, opened 2024-09-24
Adjusted test_requirements format

eddie-knight:
this changes the test_requirements object to have TLP level associated with each requirement, instead of the other way around (which resulted in repetition of test requirements)

**previous**:

```yaml
    test_requirements:
      tlp_green:
        01: |
          Simulate non-human enumeration activities (e.g., with service accounts) and verify that alerts are triggered in the logging system.
        02: |
          Confirm that logs are properly generated and accessible for review following non-human enumeration attempts.
      tlp_amber:
        01: |
          Simulate non-human enumeration activities (e.g., with service accounts) and verify that alerts are triggered in the logging system.
        02: |
          Confirm that logs are properly generated and accessible for review following non-human enumeration attempts.
```

**updated**:

```yaml
    test_requirements:
      - id: CCC.C05.TR01
        text: |
          The service blocks access to sensitive resources and admin access
          from untrusted sources, including unauthorized IP addresses, domains,
          or networks that are not included in a pre-approved allowlist.
        tlp_levels:
          - tlp_clear
          - tlp_green
          - tlp_amber
          - tlp_red
```

### DIFF
--- a/schemas/controls-schema.json
+++ b/schemas/controls-schema.json
@@ -16,7 +16,7 @@
           "properties": {
             "id": {
               "type": "string",
-              "description": "Control ID in the format CCC.<Service Category Abbreviation>.C1"
+              "description": "Control ID in the format CCC.<Service Category Abbreviation>.C##"
             },
             "title": {
               "type": "string",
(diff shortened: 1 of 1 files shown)


sshiells-scottlogic (approved):
LGTM

PR #423, opened 2024-09-24
Update controls snippet and controls and threats example files

sshiells-scottlogic:
Update the controls snippet and the threats example files to match recent changes to the schema.


eddie-knight (approved):
Thanks for staying on top of this!

PR #338, opened 2024-09-09
Setting up delivery tooling package

eddie-knight:
This is still very WIP. Been needing to import the changes from #327 and #337, so the PR looks bigger right now than it will be at the end.

New changes in this PR will be limited to the `delivery_tooling` directory.

PR #426, opened 2024-09-25
Artifact Release Pipeline

damienjburks:
Extension of #338 - still a WIP.


damienjburks:
@eddie-knight huge PR, but i think we're ready for an initial release

.gitignore
```diff
@@ -1,4 +1,8 @@
 # Name of default directory for oscal-cli
 build/oscal-cli
 # VS Code 
-.DS_Store
\ No newline at end of file
+.DS_Store
+.vscode
```
sshiells-scottlogic, line 5: Adding .vscode back in will remove the code snippets/recommended extensions. Is this intentional?

delivery-tooling/logos/sample-logo-2.svg
sshiells-scottlogic, line 1: Do we need these logos?

sshiells-scottlogic (dismissed):
Not familiar with go but looks sensible enough. I have left a couple of small comments, nothing major. Approving as I don't believe the comments should hold things up.

.gitignore
damienjburks: totally accidental. will remove

eddie-knight: Good catch. I think the real issue is that this is ineffectual because .gitignore only prevents new files from being tracked— and .vscode is already tracked.

eddie-knight (approved):
LGTM! Thanks for applying the suggested fixes

PR #445, opened 2024-10-02
October Release Candidate - Object Storage

damienjburks:
First release candidate for object storage, also, i added some automation to help out with compiling the list of changes for the change log and the authors who contributed.


damienjburks (commented):
@eddie-knight & @smendis-scottlogic  - please review/approve when you have a chance. will cut the first release candidate later today.

eddie-knight (commented):
@damienjburks I think it would be nice to have this be more concise, instead of the working-notes style of commits that we currently have.

Perhaps a future solution will be to enforce PR naming conventions, but until then I think a manual revision to the change log will be prudent (and maybe even afterward for initial releases where lots of upfront work is happening)

damienjburks:
@eddie-knight I like that idea, maybe we could introduce some additional automation with AI to help summarize the working notes and PRs in the change log. Maybe we ensure that the change log is no more than 5 lines long. Could be a great brainstorming session.

eddie-knight (approved):

smendis-scottlogic (approved):

PR #476, opened 2024-10-21
Adding In Logo Wall and Updating Release Notes Template

damienjburks:


eddie-knight:
Happy to approve this in it's current state, but I'm curious... what's the long-term plan for multiple releases? Will we always use the same logo-wall?

damienjburks:
@eddie-knight the goal i had in mind was to keep the same logo wall and update it we have more organizations represented with future releases. Sent an email out to get more opinions on this idea/approach.

eddie-knight (approved):
LGTM— can refine more based on feedback

rgriffiths-scottlogic:
Apologies for the delay - I've been on holiday!
I like the look of the wall right now - it's nice & clean :-)
I'm 100% good to approve this right now "as is" with the following thoughts:
If we're lucky enough to attract dozens more contributors from different entities, would it still look as nice/clean? Would there be any distinction between the contributors based on their actual efforts? I'm not suggesting a size-related logo! Is it worth thinking about how to recognise the early / major / significant contributors? (Another question is who decides what is "significant"?). Food for thought for when we have that nice problem to deal with!

PR #492, opened 2024-10-30
Adding TODO Checker

damienjburks:
This PR contains the addition of a new action that'll scan the YAMLs for TODO's in the comments, fail the pipeline, and pointing them out. I also fixed the linting issues for the CMB.md file.

### DIFF
--- a/.github/workflows/todo-check.yml
+++ b/.github/workflows/todo-check.yml
@@ -12,7 +12,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check for TODOs in services directory         
-        uses: damienjburks/simple-todo-checker@0.1.3-alpha
+        uses: damienjburks/simple-todo-checker@1.0.0
         with: 
           path: "./services"
     
--- a/.github/workflows/todo-check.yml
+++ b/.github/workflows/todo-check.yml
@@ -15,4 +15,7 @@ jobs:
         uses: phntmxyz/pr_todo_checker@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/todo-check.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  find-todos:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Check for TODOs in services directory         
+        uses: damienjburks/simple-todo-checker@1.0.0
```
sshiells-scottlogic, line 15: I am not familiar with the tool but always wary when I see someones idividual username in code/config.

damienjburks: ah yes - i've taken the liberty of taking a chance at creating a todo checker in the marketplace; it seemed to be a common problem that should be solved.

eddie-knight: Agree with the hesitation from @sshiells-scottlogic— but creating items in a personal namespace is a fairly common practice for small tools. As long as we maintain a policy to pin versions, we can just do a code review of the external repo before upgrading.

sshiells-scottlogic: Yeah if it is an tool that you developed then that is all good (I have missed most of the recent meetings so I might have missed some context). I was just wanting to make sure that you weren't pointing to something local while testing that would then break when merged.

sshiells-scottlogic (approved):

jared-lambert (approved):
Thanks!

PR #521, opened 2024-11-10
Adding db-user authN threat

ianwalkersmithciticom:


services/database/relational/threats.yaml
```diff
@@ -10,6 +10,7 @@ common_threats:
   - CCC.TH10 # Alerts are Intercepted
   - CCC.TH11 # Event Notifications are Incorrectly Triggered
   - CCC.TH12 # Resource constraints are exhausted
+  - CCC.TH13 # DB user authentication mechanism is misconfigured
```
mlysaght2017, line 13: @ianwalkersmithciticom - Looks good, thanks! If this a service-specific threat, then you add the threat as you did on line 78, but you don't need to it add to the list of common threats at the top of file (here on line 13). So you just need to remove CCC.TH13 on line 13

mlysaght2017 (changes requested):

services/database/relational/threats.yaml
ianwalkersmithciticom: I removed line 13... however, I didn't get your instructions - this threat is generic to cloud, not an CSP specific threat. Where should it go?

mlysaght2017 (approved):

PR #522, opened 2024-11-11
Rephrase control test requirements

eddie-knight:
This PR changes the control test requirements to follow the style of `When <Activity>, <Actor> MUST <Action>`


mlysaght2017 (approved):

eddie-knight:
Extracting the TODO note that I never managed to get past...

```
    # TODO: Should we disallow an end user interacting with their own data via API? That seems to be excluded by this definition.
    # Brainstorming... An end user might authorize program to access an API on their behalf, but it should be a known program that is
    # explicitly allowed to access API endpoints on the network. Not sure whether this covers things like Plaid, Zelle, etc.
```

mlysaght2017 (approved):

PR #526, opened 2024-11-11
Create features.yaml for Container registry

kazmik23:
still need to add reference links


eddie-knight:
@smendis-scottlogic This should be good to go

PR #529, opened 2024-11-13
Update sonatype_scan.yaml

TheJuanAndOnly99:
Uncomment line that was accidentally commented out. This will make it so it only runs the job if the repository owner is `finos`.


damienjburks (approved):

damienjburks:
thank you @TheJuanAndOnly99  - beat me to it!

Issue #533, opened 2024-11-14
Expand developers guid into contributors guide


sshiells-scottlogic:
Expand on the existing development guide to be a more general contributors guide.

Include information on repo structure, how to find features, threats, controls etc.

Expand on development section to include more details on how to use the snippets, possibly add some gifs to make it easier to see what they can do.

PR #541, opened 2024-11-15
Issue 533: Update developer documentation

dogle-scottlogic:
Update the developer document to be a more complete contributors guide.

Resolves issue #533


dogle-scottlogic:
@damienjburks - we could do, the other option is to link to this doc from the root README.md so there's just a single doc at the root level to get started with the project but happy either way?

damienjburks:
@dogle-scottlogic I think the better approach would be to move it out to the root as CONTRIBUTING.md. Let's do that.

damienjburks (dismissed):

sshiells-scottlogic (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,97 @@
...
+The resource should also contain a `tests` directory containing details on how to test the resource is compliant.
+
+The top level of the `services` directory contains `yaml` files for common controls, features and threats which are common to multiple resources. It also contains the [`service-families.yaml`](/services/service-families.yaml) file which documents the service families and their descriptions.
+
+## VSCode
+
+### Snippets
+
+[VSCode snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets) are defined for common [features](.vscode/common-features.code-snippets), [threats](.vscode/common-threats.code-snippets) and [controls](.vscode/common-controls.code-snippets).
+To make use of these snippets start typing the snippet prefix in any `yaml` file and VSCode will offer you the option to auto-complete the snippet e.g.
```
sshiells-scottlogic, line 69: Would you perhaps be able to add a grid/summary of the available snippets (general info about single common features, threats etc.).

Can be part of a different PR.

dogle-scottlogic: Just to clarify, are you asking for a high-level list of each .code-snippets file and it's purpose as opposed to a list of each individual snippet and what it's for? The latter seems like a bit of duplication and would be hard to maintain.

sshiells-scottlogic: Just a high level list e.g.

threats-file: adds the template for a threats file
threat: adds a template for a single threat
CT*: adds a common threat

Does that make sense?

I don't see us adding many more types of snippets, other than maintaining the common features, threats and controls, but those would be covered by the general explanation.

dogle-scottlogic: you mean something like [this](https://github.com/finos/common-cloud-controls/pull/541/commits/80b3253dc38bb70d7faac713eb0db96ed5585d1f)?

sshiells-scottlogic: Yup - just saves people having to go digging to see what is available.

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+| Threat            | Create a blank FINOS CCC Threat        | [threats.code-snippets](.vscode/threats.code-snippets)                 |
+| Common Threats\*  | Create a common FINOS CCC Threat       | [common-threats.code-snippets](.vscode/common-threats.code-snippets)   |
+| Features File     | Create a blank FINOS CCC Features file | [features.code-snippets](.vscode/features.code-snippets)               |
+| Feature           | Create a blank FINOS CCC Feature       | [features.code-snippets](.vscode/features.code-snippets)               |
+| Common Features\* | Create a common FINOS CCC Feature      | [common-features.code-snippets](.vscode/common-features.code-snippets) |
+| Controls File     | Create a blank FINOS CCC Controls file | [controls.code-snippets](.vscode/controls.code-snippets)               |
+| Control           | Create a blank FINOS CCC Control       | [controls.code-snippets](.vscode/controls.code-snippets)               |
+| Common Controls\* | Create a common FINOS CCC Control      | [common-controls.code-snippets](.vscode/common-controls.code-snippets) |
```
sshiells-scottlogic, line 89: We need to add the metadata option.

We should add a column with the prompts to start typing to get the snippet up?

dogle-scottlogic: I'll add the metadata one, missed that! 
For the prompts, feels like that's just duplicating the info in the linked snippet file? It's also not clear what we'd put for the more general ones e.g. `Common Features*`?

sshiells-scottlogic: it would be just so that pepole don't have to go into the files and parse the info from the snippets.

for common ones the prompt would be e.g.

cf
ct
cc

I can add these later though if I get time. It all makes sense in my head, maybe not so much when I write it down in the comments here :)

dogle-scottlogic: Ok, I've update that - I suspect our prompts for the common controls are wrong though (CT rather than CC)?

sshiells-scottlogic: That was fixed, just checking where it went...

sshiells-scottlogic: Apparently it wasn't fixed, it should have been part of this: https://github.com/finos/common-cloud-controls/commit/03264fc9f1a4b3703cbddf64dd55e33c645dd1e4

dogle-scottlogic: 😀 - I can add it in here if you like?

sshiells-scottlogic: If you have the time then yeah go for it

damienjburks (dismissed):

CONTRIBUTING.md
eddie-knight, line 1: Note that this change will also need to remove .github/CONTRIBUTING.md, which is linked under the description when folks create a PR

Issue #534, opened 2024-11-14
Add yaml formatting guidelies


sshiells-scottlogic:
Add yaml formatting guidelines to enforce consistency with things like spaces etc. (i.e. 2 v 4)

PR #542, opened 2024-11-15
#534: Add yaml linter

dogle-scottlogic:
Add [yamllint](https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration) as a workflow job to ensure the yaml files are consistent with the markdown.

Addresses issue #534 

There are quite a few fails for the linter so I'll address those in a different PR

### DIFF
--- a/.config/.markdownlint.yaml
+++ b/.config/.markdownlint.yaml
@@ -84,7 +84,7 @@ MD022:
 MD023: true
 
 # MD024/no-duplicate-heading : Multiple headings with the same content : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md024.md
-MD024: 
+MD024:
   siblings_only: true
 
 # MD025/single-title/single-h1 : Multiple top-level headings in the same document : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md025.md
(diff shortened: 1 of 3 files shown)


eddie-knight:
Looks like the linter works, and integrates with GitHub in-line comments! Nice!

Screenshot 2024-11-15 at 06 53 47

For this case, though... I noticed that this yaml isn't following the style that it enforces 😆 

Are we sure we want to force ourselves to go back and add it to every file? I could go either way— just depends on whether this is adding value.

dogle-scottlogic:
@eddie-knight 
> Are we sure we want to force ourselves to go back and add it to every file? I could go either way— just depends on whether this is adding value.

Yeah - I'm just finishing off fixing the errors on a separate branch. I've changed the settings to ignore the .config folder and extend the `relaxed` rule-set to make things a bit easier

damienjburks:
@dogle-scottlogic should we close this PR/

dogle-scottlogic:
@damienjburks - yeah we can do. I guess I had originally intended to keep adding the linter and fixing the issues as two separate commits just to keep it a bit neater but I don't think there's a huge amount of value in it as this one is pretty small. I close it down 😄.

PR #544, opened 2024-11-19
Create features.yaml for devtools/build

kazmik23:
For services such as GCP codebuild, AWS codebuild, and Azure DevOps


eddie-knight:
@smendis-scottlogic this should be good to go

smendis-scottlogic (approved):

PR #548, opened 2024-11-19
Create controls.yaml for batchproc

kazmik23:


services/compute/batchproc/controls.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - CCC.TH03 # Deployment region network is untrusted
+    nist_csf: PR.AC-5 # Network integrity is protected
+    control_mappings:
+      NIST_800_53:
+        - SC-7 # Boundary Protection
+        - SC-8 # Transmission Confidentiality and Integrity
+    test_requirements:
+      - id: CCC.BatchProc.C01.TR01
+        text: |
+          Attempt to access the compute resource over the public internet using an external IP address and verify that access is denied.
```
eddie-knight, line 28:
```suggestion
          When an external IP address attempts to access the compute resource over the public internet, the system MUST deny the request.
```

Adjusting to follow the recently proposed format of `When <activity>, <actor> MUST <action>`.

On another note... This seems like a common control for pretty much any networked red/amber resource?

kazmik23: yeah, I thought about that as well. I was thinking about how on the validation side of things, things could be different dependent on the service, so for now I left it in there, especially because there was not a direct common control for public external ip stuff right now

damienjburks:
@kazmik23 are you able to resolve the lint and yaml checks?

PR #564, opened 2024-11-26
Add in doc on alignment with FINOS AI readiness SIG

mlysaght2017:


mlysaght2017:
@eddie-knight @jared-lambert wasn't sure where to place this doc, so let me know if we need it somewhere else

eddie-knight:
@sshiells-scottlogic do you have an opinion on this file location?

sshiells-scottlogic:
@eddie-knight  yes, but what that opinion is I am not fully sure. I don't think it belongs to here for sure.

Perhaps it could be in something like: roadmap/reference-architectures ?

eddie-knight:
@sshiells-scottlogic @mlysaght2017 🤔  Would it be better to keep this info in one of the github issues related to the topic?

eddie-knight:
Alternatively, the AI Readiness SIG is working on the "AI Governance Framework" which seems like a prime candidate for the duplication reduction WG's `frameworks` directory

sshiells-scottlogic:
> @sshiells-scottlogic @mlysaght2017 🤔 Would it be better to keep this info in one of the github issues related to the topic?

If we did this there is a chance the issue is closed/deleted and we lose the info?

eddie-knight (approved):

PR #570, opened 2024-11-27
Add in fixes to control definitions

mlysaght2017:


services/common-controls.yaml
```diff
@@ -70,16 +70,18 @@ controls:
...
           - tlp_green
           - tlp_amber
           - tlp_red
       - id: CCC.C02.TR02
         text: |
-          Admin users can verify and audit encryption status for stored data at rest,
-          including verification of key management processes.
+          Admin users can verify and audit encryption status for
+          stored data at rest, including verification of key
+          management processes.
```
sarahecraddock, line 84: Do we want to make the same change here to remove admin users from description?

sarahecraddock (approved):
Looks good @mlysaght2017, just one suggestion that I spotted where the admin user language is used in a similar way that it might be worth updating to align with the others you updated 😊

services/common-controls.yaml
mlysaght2017: Done now!

mlysaght2017:
@damienjburks - can you please approve and merge - thanks

damienjburks (approved):

PR #571, opened 2024-11-28
updates to common features

ianwalkersmithciticom:
I've made some updates to features and removed those that are not common.

### DIFF
--- a/services/common-features.yaml
+++ b/services/common-features.yaml
@@ -1,13 +1,13 @@
 features:
-  - id: CCC.F01 # Encryption in Transit Enabled by Default
-    title: Encryption in Transit Enabled by Default
+  - id: CCC.F01 # Encryption in Transit
+    title: Encryption in Transit
     description: |
-      Provides default encryption of data in transit through SSL or TLS.
+      Provides encryption of data in transit through SSL or TLS.
 
   - id: CCC.F02 # Encryption at Rest Enabled by Default
-    title: Encryption at Rest Enabled by Default
+    title: Encryption at Rest
     description: |
-      Provides default encryption of data before storage, with the option for
+      Provides encryption of data in storage, with the option for
       clients to maintain control over the encryption keys.
 
   - id: CCC.F03 # Access/Activity Logs
@@ -22,23 +22,11 @@ features:
       Allows the setting of a threshold where industry-standard throughput is
       achieved up to the specified rate limit.
 
-  - id: CCC.F05 # Signed URLs
-    title: Signed URLs
-    description: |
-      Provides the ability to grant temporary or restricted access
-      to a resource through a custom URL that contains authentication information.
-
   - id: CCC.F06 # Identity Based Access Control
     title: Identity Based Access Control
     description: |
       Provides the ability to determine access to resources based on
       attributes associated with a user identity.
-
-  - id: CCC.F07 # Event Notifications
-    title: Event Notifications
-    description: |
-      Publishes events for creation, deletion, and modification of
-      objects in a way that enables users to trigger actions in response.
 
   - id: CCC.F08 # Multi-zone Deployment
     title: Multi-zone Deployment
@@ -86,12 +74,12 @@ features:
   - id: CCC.F15 # Cost Management
     title: Cost Management
     description: |
-      Provides the ability to filter spending and to detect cost anomalies for the service.
+      Provides the ability to manage costs, filter spending and to detect cost anomalies.
 
   - id: CCC.F16 # Budgeting
     title: Budgeting
     description: |
-      Provides the ability to trigger alerts when spending thresholds are approached or exceeded for the service.
+      Provides the ability to trigger alerts when spending thresholds are approached or exceeded for services.
 
   - id: CCC.F17 # Alerting
     title: Alerting


services/common-features.yaml
```diff
@@ -1,13 +1,13 @@
 features:
-  - id: CCC.F01 # Encryption in Transit Enabled by Default
-    title: Encryption in Transit Enabled by Default
```
ianwalkersmithciticom, line 3: enc in transit isn't a given or enables by default. But you can configure and use enc

services/common-features.yaml
```diff
@@ -1,13 +1,13 @@
...
     description: |
-      Provides default encryption of data in transit through SSL or TLS.
+      Provides encryption of data in transit through SSL or TLS.
 
   - id: CCC.F02 # Encryption at Rest Enabled by Default
-    title: Encryption at Rest Enabled by Default
+    title: Encryption at Rest
     description: |
-      Provides default encryption of data before storage, with the option for
+      Provides encryption of data in storage, with the option for
```
ianwalkersmithciticom, line 10: grammar change only

services/common-features.yaml
```diff
@@ -22,24 +22,12 @@ features:
       Allows the setting of a threshold where industry-standard throughput is
       achieved up to the specified rate limit.
 
-  - id: CCC.F05 # Signed URLs
```
ianwalkersmithciticom, line 25: I don't see this as a common feature. Perhaps I've misunderstood signed-urls but they're not ubiquitous.

eddie-knight: Ubiquitous and common are two different concepts— the idea is to put items in this file if they will be used by multiple other files.

For Signed URLs, we see them on at least 4 different services:

- [Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html)
- [Amazon CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html)
- [Google Cloud — Cloud Storage](https://cloud.google.com/storage/docs/access-control/signed-urls)
- [Microsoft Azure Blob Storage](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview)

On a separate note, any time we may decide to remove a common entry, we'll also need to correct places where it's already in use:

https://github.com/search?q=repo%3Afinos%2Fcommon-cloud-controls%20CCC.F05&type=code

services/common-features.yaml
eddie-knight: I was able to easily find documentation describing the places where GCP provides default encryption in transit as a key feature. If other providers aren't doing something similar already, I'd say they need to be held to the same standard.

https://cloud.google.com/docs/security/encryption-in-transit#encryption_in_transit_by_default

eddie-knight (changes requested):
Some of the proposed changes here don't seem to be accurate

services/common-features.yaml
smendis-scottlogic: If it is not enabled by default for the service that you are working on then I suggest don't include this common feature and create a specific feature for that service

services/common-features.yaml
```diff
@@ -1,13 +1,13 @@
...
-  - id: CCC.F01 # Encryption in Transit Enabled by Default
-    title: Encryption in Transit Enabled by Default
+  - id: CCC.F01 # Encryption in Transit
+    title: Encryption in Transit
     description: |
-      Provides default encryption of data in transit through SSL or TLS.
+      Provides encryption of data in transit through SSL or TLS.
 
   - id: CCC.F02 # Encryption at Rest Enabled by Default
-    title: Encryption at Rest Enabled by Default
```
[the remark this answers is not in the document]
smendis-scottlogic: If it is not enabled by default for the service that you are working on then I suggest don't include this common feature and create a specific feature for that service

services/common-features.yaml
smendis-scottlogic: Common features do not need to apply for all services. We include the common features that do apply to the specific service in the feature file. If there are more than 1 service that has the same feature or you think there are more services that can use that feature when we start working on them in future we can add it as a common feature.

services/common-features.yaml
```diff
@@ -22,24 +22,12 @@ features:
...
     title: Identity Based Access Control
     description: |
       Provides the ability to determine access to resources based on
       attributes associated with a user identity.
 
-  - id: CCC.F07 # Event Notifications
-    title: Event Notifications
-    description: |
-      Publishes events for creation, deletion, and modification of
-      objects in a way that enables users to trigger actions in response.
```
smendis-scottlogic, line 41: Not sure why this is removed

services/common-features.yaml
smendis-scottlogic: `default` should be present here

smendis-scottlogic (changes requested):

eddie-knight:
Hey @ianwalkersmithciticom, appreciate the intent here— though the changes don't seem to be aligned with the project vision.

GitHub Issues may be a good way to reopen some of this topic for consideration, or if you have any more changes that you want to find alignment on prior to working on a PR.

PR #583, opened 2024-12-08
Add in additional common threats and controls for VPC

mlysaght2017:


mlysaght2017:
@eddie-knight @damienjburks - can I get your eyes on these additional common threats and controls for vpc and then follow through with merge if all ok pls

eddie-knight:
Could you remind me why we have some situations like this where there are fewer controls than threats?

eddie-knight (approved):

damienjburks (approved):

PR #640, opened 2025-01-20
Added missing mappings to NIST controls

abikhuil:
Added missing mappings to NIST controls for common controls. Does not include CCM/ISO mappings

### DIFF
--- a/services/common-controls.yaml
+++ b/services/common-controls.yaml
@@ -57,7 +57,9 @@ controls:
     control_mappings:
       CCM: []
       ISO_27001: []
-      NIST_800_53: []
+      NIST_800_53:
+        - SC-13 # Use of cryptographic controls
+        - SC-28 # Protection of information at rest
     test_requirements:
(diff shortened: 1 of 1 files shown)


services/common-controls.yaml
```diff
@@ -160,7 +162,10 @@ controls:
     control_mappings:
       CCM: []
       ISO_27001: []
-      NIST_800_53: []
+      NIST_800_53:
```
eddie-knight, line 165:
```suggestion
      NIST_800_53:
```

services/common-controls.yaml
```diff
@@ -418,7 +427,8 @@ controls:
...
```
eddie-knight, line 430:
```suggestion
      NIST_800_53:
```

eddie-knight (commented):
Awesome! Thanks for adding these @abikhuil

mlysaght2017 (approved):
LGTM - Thanks @abikhuil !

PR #663, opened 2025-02-08
Stylesheet Fix

robmoffat:
Fixed missing stylesheet issue


robmoffat:
Hi @sshiells-scottlogic - can you take a look at the preview and let me know if this works for you?  I'm fixing an issue with a missing stylesheet and also reducing the header size which Eddie complained about (I think he was right).

@Alexstpierrework fyi

sshiells-scottlogic (approved):

PR #676, opened 2025-02-24
Threats reviewed and amended for RDMS

abikhuil:
Made the following amendments 
- Added missing common feature to feature list
- Amended ID ordering
- Removed duplicated threats
- Added new common threats to RDMS

### DIFF
--- a/services/database/relational/threats.yaml
+++ b/services/database/relational/threats.yaml
@@ -10,23 +10,11 @@ common_threats:
   - CCC.TH10 # Alerts are Intercepted
   - CCC.TH11 # Event Notifications are Incorrectly Triggered
   - CCC.TH12 # Resource constraints are exhausted
+  - CCC.TH13 # Resource tags are manipulated
+  - CCC.TH15 # Automated enumeration and reconnaissance by non-human entities
 
 threats:
   - id: CCC.RDMS.TH01
-    title: Unauthorized Access to Database
-    description: |
-      Unauthorized access to a cloud relational database may
-      occur due to the use of compromised roles or default
-      administrative credentials. This condition could result in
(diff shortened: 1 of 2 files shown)


services/database/relational/threats.yaml
```diff
@@ -70,21 +58,7 @@ threats:
...
-      through network eavesdropping.
-    features:
-      - CCC.F01 # Encryption in Transit
-      - CCC.F02 # Encryption at Rest
-    mitre_technique:
-      - TA0006
-      - T1040
-
-  - id: CCC.RDMS.TH06
+  - id: CCC.RDMS.TH04
```
mlysaght2017, line 61: @abikhuil @eddie-knight - This threat as its written sounds a little vague in terms of its impact. Would like to get your thoughts on how we can make it a little more clear - e.g. does it relate to [Data Encrypted for Impact](https://attack.mitre.org/techniques/T1486/) or is it more about the loss of confidentiality?

mlysaght2017 (changes requested):
Overall, LGTM - thanks @abikhuil. Left a comment for a specific threat. 

@abikhuil - are you ok to cover RDMS controls next? Thanks!

services/database/relational/threats.yaml
eddie-knight: If we're talking about `Snapshot Collection with Unauthorized Encryption Key`... that whole entry is confusing for me at the moment. I feel like there must be something technical I'm misunderstanding 🤔

mlysaght2017: @zigmax - interested to get your views here and your feedback on the full threat list for RDMS please - thanks

mlysaght2017 (approved):

PR #699, opened 2025-03-20
Applied schema changes to ObjStor for feedback

eddie-knight:
This is a preview of the changes that would be applied by the recent round of feedback. I've written a parser to automate the conversion once we're happy with the target data structure. The same code can be referenced to update the delivery tooling to keep that part in sync.

Pending questions:

- Do we like the topical segmentation provided by nesting controls within control families?
  - Pro: Shortcuts topical sorting for human- and machine-readability. Highlights topical areas that are not yet covered.
  - Con: _Technically_ superlative to simply putting the control family as a field.
- Do we want to add categories or families to threats and capabilities/features?
- If yes to both of the previous questions... do we want to handle all of the data types with similar nesting?

### DIFF
--- a/services/storage/object/capabilities.yaml
+++ b/services/storage/object/capabilities.yaml
@@ -1,85 +1,99 @@
-common_features:
-  - CCC.F01 # Encryption in Transit Enabled by Default
-  - CCC.F02 # Encryption at Rest Enabled by Default
-  - CCC.F03 # Access/Activity Logs
-  - CCC.F04 # Transaction Rate Limits
-  - CCC.F05 # Signed URLs
-  - CCC.F06 # Identity Based Access Control
-  - CCC.F07 # Event Notifications
-  - CCC.F08 # Multi-zone Deployment
-  - CCC.F09 # Monitoring
-  - CCC.F10 # Logging
-  - CCC.F11 # Backup
-  - CCC.F12 # Restore
-  - CCC.F14 # API Access
-  - CCC.F18 # Versioning
-  - CCC.F20 # Tagging
-  - CCC.F21 # Replication
-  - CCC.F22 # Location Lock-In
+shared-capabilities:
+  - reference-id: CCC
+    identifiers:
+      - CCC.F01 # Encryption in Transit Enabled by Default
+      - CCC.F02 # Encryption at Rest Enabled by Default
+      - CCC.F03 # Access/Activity Logs
+      - CCC.F04 # Transaction Rate Limits
+      - CCC.F05 # Signed URLs
+      - CCC.F06 # Identity Based Access Control
+      - CCC.F07 # Event Notifications
+      - CCC.F08 # Multi-zone Deployment
+      - CCC.F09 # Monitoring
+      - CCC.F10 # Logging
+      - CCC.F11 # Backup
+      - CCC.F12 # Restore
+      - CCC.F14 # API Access
+      - CCC.F18 # Versioning
+      - CCC.F20 # Tagging
+      - CCC.F21 # Replication
+      - CCC.F22 # Location Lock-In
 
-features:
+capabilities:
   - id: CCC.ObjStor.F01
     title: Storage Buckets
     description: |
       Provides uniquely identifiable segmentations in which data elements may
       be stored.
+    category: ""
 
   - id: CCC.ObjStor.F02
     title: Storage Objects
     description: |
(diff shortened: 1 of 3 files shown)


services/storage/object/capabilities.yaml
```diff
@@ -1,85 +1,99 @@
...
-  - CCC.F10 # Logging
-  - CCC.F11 # Backup
-  - CCC.F12 # Restore
-  - CCC.F14 # API Access
-  - CCC.F18 # Versioning
-  - CCC.F20 # Tagging
-  - CCC.F21 # Replication
-  - CCC.F22 # Location Lock-In
+shared-capabilities:
+  - reference-id: CCC
```
eddie-knight, line 2: > [!NOTE]
> We'll likely want to version-control the shared entries separately, and reference by version here.

services/storage/object/capabilities.yaml
```diff
@@ -1,85 +1,99 @@
...
+      - CCC.F22 # Location Lock-In
 
-features:
+capabilities:
   - id: CCC.ObjStor.F01
     title: Storage Buckets
     description: |
       Provides uniquely identifiable segmentations in which data elements may
       be stored.
+    category: ""
```
eddie-knight, line 28: > [!NOTE]
> This is what it would look like if we add categories (or families) but do not add nesting

services/storage/object/controls.yaml
```diff
@@ -1,276 +1,268 @@
...
-      Ensure that uniform bucket-level access is enforced across all
-      object storage buckets. This prevents the use of ad-hoc or
-      inconsistent object-level permissions, ensuring centralized,
-      consistent, and secure access management in accordance with the
-      principle of least privilege.
-    control_family: Identity and Access Management
-    nist_csf: PR.AC-4 # Access permissions and authorizations are managed,
-    # incorporating the principles of least privilege and separation of duties
-    threats:
-      - CCC.TH01 # Access Control is Misconfigured
```
eddie-knight, line 82: > [!WARNING]
> My parser isn't preserving in-line comments, and I'm not sure I'll get it to do so successfully. Any of these that are retained are being copied manually.

services/storage/object/controls.yaml
```diff
@@ -1,276 +1,268 @@
...
-      - id: CCC.ObjStor.C06.TR01
-        text: |
-          When an object storage bucket is accessed, the service MUST store
-          access logs in a separate data store.
-        tlp_levels:
-          - tlp_amber
-          - tlp_red
+control-families:
+  - name: Data
+    description: ""
```
eddie-knight, line 16: > [!NOTE]
> We do not currently have definitions or descriptions for control families. If we like this structure, we can put that information here. Otherwise, we'll want to store it elsewhere — probably in the metadata.

services/storage/object/controls.yaml
```diff
@@ -1,276 +1,268 @@
...
+        requirements:
+          - id: CCC.ObjStor.C01.TR01
+            text: |
+              When a request is made to read a protected bucket, the service
+              MUST prevent any request using KMS keys not listed as trusted by
+              the organization.
+            applicability:
+              - tlp_amber
+              - tlp_red
+            recommendation: ""
```
eddie-knight, line 33: > [!NOTE]
> If we like this, the `recommendation` field is where we can (optionally) leave suggestions for implementing this control.

services/storage/object/capabilities.yaml
```diff
@@ -1,85 +1,99 @@
...
-  - CCC.F09 # Monitoring
-  - CCC.F10 # Logging
-  - CCC.F11 # Backup
-  - CCC.F12 # Restore
-  - CCC.F14 # API Access
-  - CCC.F18 # Versioning
-  - CCC.F20 # Tagging
-  - CCC.F21 # Replication
-  - CCC.F22 # Location Lock-In
+shared-capabilities:
```
robmoffat, line 1: Didn't we decide in a previous meeting to not rename features to capabilities?

robmoffat: Would like to see mappings for controls at least moved into another file.   There are various pros and cons of this:

1.  Mappings will become neglected if we separate them. 
2. On the other hand, it frees up the security team to not have to worry so much about these.
3. We may be able to generate control mappings using tools
4. We may be able to more easily version control mappings on a per-file basis.  e.g. we can have a separate file for each version of a control catalog./

Issue #693, opened 2025-03-11
Architectural Components for the Three Tier App


robmoffat:
From here: 

https://cloud.google.com/architecture/single-zone-deployment-compute-engine

https://cloud.google.com/application-design-center/docs/three-tier-web-app#:~:text=This%20guide%20helps%20you%20understand%20the%20Three%20tier,in%20the%20middle%20tier.%20A%20SQL%20database%20backend

From the first link:  

This reference architecture uses the following Google Cloud products:

- [Compute Engine](https://cloud.google.com/compute): A secure and customizable compute service that lets you create and run VMs on Google's infrastructure.
  - [x] Features Defined
  - [x] Reviewed
- [Cloud Load Balancing](https://cloud.google.com/load-balancing): A portfolio of high performance, scalable, global and regional load balancers.
  - [x] Features Defined
  - [ ] Reviewed
  - [ ] Controls Defined
  - [ ] Released
- [Cloud Storage](https://cloud.google.com/storage): A low-cost, no-limit object store for diverse data types. Data can be accessed from within and outside Google Cloud, and it's replicated across locations for redundancy.\
  - [x] Features Defined
  - [x] Reviewed
  - [x] Controls Defined
  - [x] Released 
- [Virtual Private Cloud (VPC)](https://cloud.google.com/vpc): A virtual system that provides global, scalable networking functionality for your Google Cloud workloads. VPC includes VPC Network Peering, Private Service Connect, private services access, and Shared VPC.
  - [x] Features Defined
  - [x] Reviewed
  - [x] Controls Defined
  - [x] Released 
- IAM
  - [ ] Features Defined - #636 
  - [ ] Reviewed
  - [ ] Controls Defined
  - [ ] Released

PR #720, opened 2025-04-17
Add in service mappings for three tier app

mlysaght2017:
So that we can reference this via a dedicated doc...

I think we really do need to implement this reference architecture for at least one cloud provider—preferably all three—to ensure that we are capturing the right set of foundational services and configuration patterns. 

While the [Google Cloud reference document](https://cloud.google.com/architecture/single-zone-deployment-compute-engine) scopes this architecture to a small set of services (primarily four), I reckon that limiting ourselves to a theoretical mapping of just 4 services will leave us with a pretty big validation gap. This feels like a good early opportunity to do that validation. 

You'll see in the table that I've added additional Sec/Mangement/Obs/Crypto services that I believe the community would be expecting to be in place in a realistic implementation. 

One other concern... I honestly don't believe any FSI would build an app on a VM today...Should we swap out for Serverless?


eddie-knight:
this relates to #693

I believe we decided yesterday that swapping to serverless is prudent, based on FSI feedback.

eddie-knight (approved):

PR #733, opened 2025-04-27
feat: Update validation to use validate against latest schema

sshiells-scottlogic:


.vscode/settings.json
sshiells-scottlogic, line 1: @eddie-knight  just an FYI - this started to flag validation issues in the UI for in the common files (which have a different file structure).

sshiells-scottlogic:
@eddie-knight  I think this should have the inline comment identifiers back in place. If you spot any that are missing then there is something wrong with the new validation code.

There are some errors left, somehow ai-ml/mlde has controls for mlde threats, but no threats in the file?

image


image

eddie-knight:
Yeah, that's a long-standing error — it seems we merged in the controls without the accompanying threats a while back. See the files in the commit prior to this merge:

https://github.com/finos/common-cloud-controls/tree/9e049c42aa08a0a313a47aa65cb68c84c3b04998/services/ai-ml/mlde

sshiells-scottlogic:
So how do we fix this? For me it doesn't really make sense (in the context of the project) to have controls for threats that don't exist?

@mlysaght2017 any thoughts?

services/crypto/key/controls.yaml
```diff
@@ -69,7 +69,8 @@ control-families:
         threat-mappings:
           - reference-id: CCC
             identifiers:
-              - CCC.KeyMgmt.TH01 # Deletion or disabling of KMS key versions leading to denial of service or data destruction
```
sshiells-scottlogic: @eddie-knight  does this look like a reasonable way to split long comments to pass linting? I have used the '|' as a separator so the validator knows to check the next line.

sshiells-scottlogic:
> Yeah, that's a long-standing error — it seems we merged in the controls without the accompanying threats a while back. See the files in the commit prior to this merge:
> 
> https://github.com/finos/common-cloud-controls/tree/9e049c42aa08a0a313a47aa65cb68c84c3b04998/services/ai-ml/mlde

Just an FYI we discussed this on the call and we decided to flag these as warnings as they are not a priority to fix so I have adjusted the validatior to handle this case.

services/crypto/key/controls.yaml
eddie-knight: Reasonable, absolutely.

Though — if we can manage it — I think I'd prefer to enforce a max character length for titles

sshiells-scottlogic: What length would you like and can you suggest an alternative title for this one and I'll try to make it happen.

I'd likely do that as a separate PR just to keep things cleaner.

eddie-knight: I don't recall seeing the error this was trying to solve, but I suppose it'd be something beneath that

PR #750, opened 2025-05-18
add(arch): catalog for 3-tier web app architecture

eddie-knight:
This PR attempts to demonstrate what it could look like to draw elements from service-specific control catalogs into a control catalog for a reference architecture.


eddie-knight:
@mlysaght2017 What do you think of this approach?

architectures/README.md
```diff
@@ -0,0 +1,13 @@
+# CCC Reference Architectures
+
+This directory contains control catalogs that are designed for complex inter-networked systems.
+
+The intent and format of these documents contrasts with the other parts of CCC in approach, but adheres to the same schema.
+
+However, architectures will not always have capabilities or threats documented, as those are likely to be inherited by the shared control catalogs. In the event that additional threats are found, they should have corresponding controls captured within the `control-families` field.
```
mlysaght2017, line 7: Where will the additional threats themselves be captured? I think we should have a threats.yaml as well as it maintains the symmetry and if we're pulling in specific controls from the shared catalog, we're implicitly doing the same for threats, right?

architectures/three-tier-web-app/metadata.yaml
```diff
@@ -0,0 +1,50 @@
+metadata:
+  id: CCC.RDMS
+  title: CCC Relational Database Management System Capabilities
+  description: |
+      This documents the minimual set of capabilities that should be present for a RDMS
+      service to be considered for use in financial services ecosystems.
+  version: "2025.05"
+  last-modified: "2025-05-18T09:14:00-06:00"
+  applicability-categories: []
```
mlysaght2017, line 9: what are you thinking here?

mlysaght2017 (commented):
Looks good!

architectures/README.md
eddie-knight: Ah, yep - makes sense

damienjburks (approved):
This is so dope! Love it 🔥

PR #768, opened 2025-06-05
Rename features to capabilities in Load Balancer

smendis-scottlogic:


eddie-knight:
Some additional changes were required for this, which I've taken care of in #766

PR #793, opened 2025-06-25
add(capability): containers

pedrooot:


smendis-scottlogic:
@pedrooot I did some digging into the 3 services we are comparing here under containers. ECS is a fully managed container orchestration platform. ACI and Cloud Run looks like just serverless containers. And sadly I can't find a native equivalent to ECS in Azure and GCP. Is it ok if I combine the work you & I did and come up with a single category called Container Orchestrator.

PR #803, opened 2025-07-01
change(ci): simplify codeowners

eddie-knight:
The comments in this file have increased maintenance overhead, so this change removes them.

It also removes the Delivery WG from codeowners ahead of the official decommissioning of that group, and makes some other changes to required reviewers.

I recommend looking at the final changed file instead of the diff, which is a bit cumbersome.


rgriffiths-scottlogic (approved):
LGTM - approved!

PR #825, opened 2025-07-14
add(threat): Add Logging threats

gabriellavengeo:
Work done mostly by @yfolias

### DIFF
--- a/services/management/logging/threats.yaml
+++ b/services/management/logging/threats.yaml
@@ -1,0 +1,133 @@
+shared-threats:
+  - reference-id: CCC
+    identifiers:
+      - CCC.TH01 # Access Control is Misconfigured
+      - CCC.TH02 # Data is Intercepted in Transit
+      - CCC.TH03 # Deployment Region Network is Untrusted
+      - CCC.TH04 # Data is Replicated to Untrusted or External Locations
+      - CCC.TH05 # Data is Corrupted During Replication
+      - CCC.TH06 # Data is Lost or Corrupted
+      - CCC.TH07 # Logs are Tampered With or Deleted
+      - CCC.TH09 # Logs or Monitoring Data are Read by Unauthorized Users
+      - CCC.TH10 # Alerts are Intercepted
+      - CCC.TH11 # Event Notifications are Incorrectly Triggered
+      - CCC.TH12 # Resource Constraints are Exhausted
+      - CCC.TH14 # Older Resource Versions are Exploited
+      - CCC.TH15 # Automated Enumeration and Reconnaissance by Non-Human Entities
+      - CCC.TH16 # Logging and Monitoring are Disabled
+      - CCC.TH17 # Unauthorized Network Access via Misconfigured Rules
+
+threats:
+  - id: CCC.Logging.TH01
+    title: Log Ingestion Performance Degradation
+    description: |
+      The logging service's ingestion pipeline experiences performance
+      degradation due to overwhelming log volumes, network bottlenecks, or
+      inefficient processing, leading to delayed availability of log data for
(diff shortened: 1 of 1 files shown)


services/management/logging/threats.yaml
```diff
@@ -0,0 +1,159 @@
...
+          - A09:2021
+      - reference-id: CWE
+        identifiers:
+          - CWE-79
+          - CWE-117
+          - CWE-116
+  - id: CCC.Logging.TH07
+    title: Insufficient Logging
+    description: |
+      If security critical actions are not logged then it increases the difficulty to
```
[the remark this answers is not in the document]
gabriellavengeo: updated

eddie-knight:
It seems that co-authoring is not allowed by EasyCLA. We'll have to decide whether to adjust the commit history here, or bypass the failing check.

https://docs.linuxfoundation.org/lfx/easycla/v2-current/easycla-and-co-author-compliance-guide#why-co-authors-are-not-supported

services/management/logging/threats.yaml
```diff
@@ -0,0 +1,159 @@
...
+          - CCC.Logging.F03 # Real-Time Log Ingestion
+          - CCC.Logging.F04 # Centralised Log Collection
+          - CCC.Logging.F06 # Log Filtering & Transformation
+    external-mappings:
+      - reference-id: MITRE-ATT&CK
+        identifiers:
+          - T1499 # Endpoint Denial of Service
+          - T1498 # Network Denial of Service
+          - T1562.008 # Impair Defenses: Disable Cloud Logs
+  - id: CCC.Logging.TH02
```
mlysaght2017, line 41: This needs some work: The title and description suggest log data exfiltration, but the mechanism described is closer to abuse of legitimate access, not exfiltration in the classic sense (e.g., unauthorized data transfer out of a trusted boundary).

services/management/logging/threats.yaml
```diff
@@ -0,0 +1,159 @@
...
+      - reference-id: CCC.Logging
+        identifiers:
+          - CCC.F22 # Location Lock-In
+    external-mappings:
+      - reference-id: MITRE-ATT&CK
+        identifiers:
+          - T1048 # Exfiltration Over Alternative Protocol
+          - T1567 # Exfiltration Over Web Service
+          - T1020 # Automated Exfiltration
+  - id: CCC.Logging.TH03
```
mlysaght2017, line 62: I'm interested to get thoughts on whether this is an operational risk as opposed to a cyber threat - and if the former, do we want to cover in CCC? While we we've steered away from using "adversary attack" phrasing in the definition of our threats, we have had the broad goal of mapping to MITRE ATT&CK TTPs. If the identified threat doesn't really map to an adversary tactic/technique identified by MITRE in the true sense (i.e. an attacker would actually use the "procedure") - then I think we should consider not including it... @eddie-knight  @zigmax

mlysaght2017 (changes requested):

services/management/logging/threats.yaml
eddie-knight: I don't know... I kindof like this one. I'm not sure if it's impractical or clever or both, but I could imagine a lot of other mitigations would be nullified if some malicious or incompetent insider were able to modify a custom schema that log/mon/alert systems are relying on.

@gabriellavengeo can you help us understand how this threat could manifest in a real world scenario? Is it *possible* for someone to destructively modify a schema either accidentally or intentionally?

TheJuanAndOnly99:
Hi @robmoffat as Eddie [mentioned above](https://github.com/finos/common-cloud-controls/pull/825#issuecomment-3108735184) EasyCLA does not support co-authoring. I would suggest to adjust the commit history if possible.

PR #843, opened 2025-07-24
fix(arch): 3-tier-web-app

smendis-scottlogic:
@gabriellavengeo noted that we have not done the Web Application Firewall part implemented. @robmoffat will be doing the tracing service. Do you think you will have capacity to do any contribution in the WAF area? Thanks!

### DIFF
--- a/architectures/three-tier-web-app/README.md
+++ b/architectures/three-tier-web-app/README.md
@@ -36,11 +36,11 @@ This document maps the services used in a standard 3-tier application architectu
 
 ## 📊 Monitoring and Logging
 
-| **Component**     | **Google Cloud**        | **Amazon Web Services (AWS)** | **Microsoft Azure**                | **CCC Service**         |
-| ----------------- | ----------------------- | ----------------------------- | ---------------------------------- | ----------------------- |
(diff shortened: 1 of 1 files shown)


rkenefeck:
@smendis-scottlogic I think Firewall Rules are still a valid element in here, as they align with the 3 Tier architecture concerns of Internal Loadbalancing. WAF is added at the External Loadbalancing component, and is already partially captured with [CCC.LB.F23](https://github.com/finos/common-cloud-controls/blob/0c7ba6355f0ccb3d2b83c578f607501332f16cb3/services/networking/loadbalancer/capabilities.yaml#L139)  Firewall Integration. 

Google's diagram shows the WAF implementation supports this idea with the Cloud Armor service hanging off External Loadbalancers https://storage.googleapis.com/gweb-cloudblog-publish/images/4_cloud_armor_ddos.max-1100x1100.jpg [Blog Article](https://cloud.google.com/blog/products/identity-security/new-waf-capabilities-in-cloud-armor)

I think its great to expand the coverage with WAF but not at the expense of Firewall Rules

PR #845, opened 2025-07-29
add(ci): Create PR with compiled yaml asset into website directory

eddie-knight:
This extends the release pipeline so that the compiled YAML asset is PR'd into the `website/static/releases` directory. This is a prerequisite for automating the display of releases on the website.

Example execution:

https://github.com/eddie-knight/common-cloud-controls/pull/2

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,8 +17,11 @@ jobs:
       run:
         working-directory: ./delivery-toolkit
     steps:
-      - uses: actions/checkout@v4
-        name: Build
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -67,6 +70,39 @@ jobs:
           if-no-files-found: error
           retention-days: 1 # Maximum Retention
 
+      - name: Copy YAML artifact to website/static/releases
```
robmoffat, line 73: You could do it this way, but it would be better to put the releases into website/src/data/ccc-releases rather than website/static as this is for serving static files out of Docusaurus.  

Also, I wrote a script website/src/scripts/DownloadCCCReleases.ts which runs when docusaurus gets built (locally or on netlify) which will update the releases in the directory anyway, so we probably don't really need to create new PRs for this

PR #857, opened 2025-08-11
add(category): Upgrade schema to OpenSSF Gemara Layer 2 (latest)

eddie-knight:
This will set us up to version control our common entries, instead of releasing common entries nested within each catalog.

After this PR merges, I will create a standalone release of the core control catalog (these have already been through the CMB) and then I'll revisit the import strategy in other catalogs prior to their re/release.


sshiells-scottlogic:
I am not 100% sure what you mean but possibly the patterns?

image

rgriffiths-scottlogic (dismissed):
I like this idea very much @eddie-knight - it is very sensible!
I trust that the schema definitions are correct (and will let someone else check them!) but I wholeheartedly endorse the notion.
Approved!

core/README.md
```diff
@@ -0,0 +1,71 @@
+# Core Reusable Catalog
+
+This catalog is not intended for end user adoption, but rather acts as a central mechanism for all other catalogs to reference. By maintaining this as a standalone version-controlled catalog, we can increase clarity as to the source of the controls when elements from this directory are brought in as shared entries to service-specific catalogs.
+
+This catalog may be imported ot other catalogs by adding the catalog data in a mapping reference entry, then targeting the relevant elements each by ID.
```
sshiells-scottlogic, line 5:
```suggestion
This catalog may be imported to other catalogs by adding the catalog data in a mapping reference entry, then targeting the relevant elements each by ID.
```

core/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+```yaml
+metadata:
+  id: NewSvc
+  title: Some New Cloud Service Category
+  version: ""
+  description: ""
+  last-modified: ""
+  mapping-references:
+    - id: CCC
```
sshiells-scottlogic, line 21: Should we have a separate id for the core catalog to make it clearer when we are referencing something from the core catalog v something for the service catalog(s)

sshiells-scottlogic (dismissed):
Approved with a couple of minor comments.

core/README.md
eddie-knight: We could do `CCC.Core`, WDYT?

eddie-knight:
This only just occurred to me, but... we have a CI check to flag TODOs, and it definitely isn't working. There are 24 cases of `description: 'TODO: Describe this control family'`.

sshiells-scottlogic:
I will try to get through this before #2 arrives.

It's not that big, still 121 files changed short of one I reviewed last week...

mlysaght2017 (approved):

rgriffiths-scottlogic (dismissed):
That sure is a lot of files! I can't say that I reviewed every single change... but I could spot the patterns repeating a lot so made the logical leap that all such changes would be fine...

PR #859, opened 2025-08-16
change(ci): Revise release compiler

eddie-knight:
Wrapping up this PR now, need to fix the go md template and populate the following change log.

## What you'll see in this PR

1. No longer release the markdown file, only PDF.
    - This is because the requirements for the PDF styling have become complex enough that it isn't prudent to try maintaining them in a way that will also properly manifest on something like GitHub. And since we don't have use-case examples of the markdown being used, I didn't bother creating a second template for us to maintain.
2. Participants are now tracked in a YAML file instead of a markdown file
    - This enables more precision when correlating commits to contributors in the release details
3. File ingestion is now done by the OpenSSF Gemara go library
    - Release details are an exception. Gemara does not currently allow for extensible objects, so release details need to be handled with custom logic. See the following point.
4. Release details are now in a standalone file.
    - See previous point. Also, note that the commit details are managed by the delivery toolkit, which should reduce maintenance complexity somewhat.
5. Core controls are no longer elucidated within the compiled release YAML, only IDs.
    - It will be up to the catalog metadata to specify which version of the core catalog is being referenced, and then an ingestion point such as the website can choose how to render that information. Data quality must be impeccable for this to work.
7. `sample-artifacts/` to demonstrate what the compiled yaml and md files look like

## Some things that weren't addressed here

1. The toolkit flags and args are different from one command to the next
1. It will be prudent to refactor the latestReleaseDetails logic so that it has less duplication with the Metadata file
1. The delivery toolkit still needs unit tests, not just for dogma but that activity will help improve the quality overall
1. No catalogs were updated to make them ready for release. The storage/object catalog was updated for local testing, but is not confirmed for a release
1. Regarding the previous two points... integration tests or makefile test runs would help increase future stability
1. We can (and probably should) drop a lot of the release details values that are unused. Many of these were put in place for potentially interested contributors, but have remained unused across the board.


robmoffat:
My preference would be to include core controls within the release too

catalogs/storage/object/release-details.yaml
```diff
@@ -0,0 +1,105 @@
...
+    - Added CCC.OS.C6 Control (#298)Looks good
+    - Extend Object Storage Controls (#263)
+  contributors:
+    - name: Damien Burks
+      github-id: damienjburks
+      company: Citi
+    - name: Eddie Knight
+      github-id: eddie-knight
+      company: Sonatype
+    - name: sshiells-scottlogic
```
rgriffiths-scottlogic, line 50: I think this is where the name/github bug might be located @eddie-knight

eddie-knight: on it, thanks for catching

eddie-knight:
The prettier check was passing locally and failing in CI, even after updating all NPM dependencies locally, so I've removed it. We have enough other linters anyhow.

This PR is ready for review.

mlysaght2017 (approved):

rgriffiths-scottlogic (approved):
Approved

rgriffiths-scottlogic (approved):
Approving with my WG Community Structure hat on

PR #864, opened 2025-08-24
Add 8 letter limitation as per Threat schema

### LINKED ISSUE #830: Threat Id's have a different validation schema than other Id's
### Description of Problem:
 The schema for Threat ID's (https://github.com/finos/common-cloud-controls/blob/e03f6caec021889da90983541d65c5d0671f5faf/schemas/threats-schema.json#L35) has an 8 character limit, but this limit doesn't exist in Capabilities or Controls (https://github.com/finos/common-cloud-controls/blob/e03f6caec021889da90983541d65c5d0671f5faf/schemas/controls-schema.json#L47)

### Potential Solutions:
I think this needs to be consistently handled between the schemas, if there is a limitation from the threat file for 8 chars or less then the same limitation should be included in the other schemas. Conversely, if there is no need for an 8 char limit, then aligning the threat schema would be the right step forward.

rkenefeck:
Fixes #830


rkenefeck:
The failing check is due to other files that were not prettier formatted. I've resolved those in PR #866 - I can rebase and force push if we cant bypass the check.

zigmax (approved):
Looks good to me. Thanks.

PR #871, opened 2025-09-09
fix(ci): updated release artifact to be a top-level catalog

eddie-knight:
This adds a single go command to consoldate the`yaml`, `md`, and `release-notes` commands into one `generate-release-artifacts`, and adjusts the releases back to a state which does not slap the release-details parallel to the catalog, so that catalogs can be ingested directly by automated tools without knowledge of the custom structure.


eddie-knight:
Hey doctor @rgriffiths-scottlogic, could you punch this Delivery PR through

rgriffiths-scottlogic (approved):
Approved

PR #879, opened 2025-09-15
fix(ci): generate release-details.yaml

eddie-knight:
thanks @robmoffat for catching this bug


rgriffiths-scottlogic (approved):
Stevie can fix his own name later!

PR #892, opened 2025-09-30
add(threat): Added capability and threat for Active Ingestion

eddie-knight:
This capability is for any service that takes inputs while running (pretty much everything).

If it takes input, there's a chance it could be spammed and DOS the service or system or network.


smendis-scottlogic (approved):
Looks good from Taxonomy point of view

mlysaght2017 (approved):

PR #897, opened 2025-10-01
WIP: Core controls wording review

sublimino:
WIP

### DIFF
--- a/catalogs/core/ccc/controls.yaml
+++ b/catalogs/core/ccc/controls.yaml
@@ -1,47 +1,40 @@
 control-families:
   - title: Data
     description: |
-      The Data control family ensures the confidentiality, integrity,
-      availability, and sovereignty of data across its lifecycle.
-      These controls govern how data is transmitted, stored,
-      replicated, and protected from unauthorized access, tampering,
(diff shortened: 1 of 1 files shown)


eddie-knight:
Here is the style guide we wrote a while back... @sublimino we discussed the validity of this yesterday, but — @mlysaght2017's point — if we pivot the style, we'll need to update the style guide accordingly.

sublimino:
Are they written in Gherkin to facilitate security automation, or as a stylistic choice?

The adherence to sentence composition makes some of them unwieldy for a human to read, which is the the energy of most edits here. Happy to discuss.

mlysaght2017:
I believe we initially wrote up the requirements in this Gherkin-like syntax because it provided a clear way of expressing the “conditions and outcomes” that the CFI implementation team needed to validate. The intent was to bridge requirements with test execution by showing precisely what a test should seek to prove. That said, I agree that while this approach may be useful for those writing or automating the tests, the syntax can be tricky to follow for readers who only care about the requirements themselves rather than the test logic. 

I support the change proposed here. 

@eddie-knight - is the CFI losing any significant benefit if the gherkin like style is dropped?

PR #898, opened 2025-10-02
change(capability): Remove strength in capabilities & removal of service-categories file

smendis-scottlogic:


eddie-knight:
1. `comparable-csp-services` is more understandable... though I think a barebones name might be easier... maybe something like `example-services`? 
2. Potentially overthinking this... Do we need to put a version identifier or date on there? I'm imagining a scenario where a service might change capabilities over time and become a less good example but older versions of the catalog are still referencing it.

eddie-knight:
Cool, I'm happy with pushing it through like that. Thanks @smendis-scottlogic

PR #921, opened 2025-10-16
change(ci): Prepare Object Storage for new release

eddie-knight:
This PR does some basic cleanup on the ObjStor catalog and refines the mappings in the control catalog.

It also adds logic to the CI to pull in the CCC.Core data from the release artifact. And the CI now also allows catalog control families to have their descriptions auto-populated at build time if they match previously used family names. For starters, this queues up `Data` and `IAM` descriptions.

There is a bit more work to do before the ObjStor release. I'm calling it a night, and opening this PR for an iterative review.

### DIFF
--- a/catalogs/storage/object/capabilities.yaml
+++ b/catalogs/storage/object/capabilities.yaml
@@ -78,7 +78,8 @@ capabilities:
   - id: CCC.ObjStor.CP08
     title: Lifecycle Policies
     description: |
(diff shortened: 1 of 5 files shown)


smendis-scottlogic:
I have updated the core capabilities of object storage in this PR #922 which needs to be merged with this PR before cutting the release for the object storage @eddie-knight

mlysaght2017 (changes requested):

PR #922, opened 2025-10-16
change(capability): Update the object storage capabilities

smendis-scottlogic:
Revise and update the capabilities of the object store for the upcoming release with new core capabilities


eddie-knight:
I'm confused by the proposed changes here — could you update the description with the rationale?

smendis-scottlogic:
Active ingestion I added back since technically you can do DDoS on static websites hosted on S3

PR #960, opened 2026-02-18
add(capability): Managed K8s as a service

smendis-scottlogic:
We have previously decided not to have native container orchestration as a service as it does not exist in Azure and Google Cloud. We decided to have managed Kubernetes as a service. This PR is not to list out the capabilities of K8s itself, but to identify the features of the wrapper around K8s provided by all service providers.


eddie-knight (approved):
We will want to get more reviews before a release is made, but I like this as an iteration. Thanks for pushing this forward @smendis-scottlogic

Issue #993, opened 2026-05-18
Release Action


robmoffat:
### Description of Problem:

We need to simplify and streamline the CCC release process - we haven't done any releases in a long time and it's over-complicated.  

### To Do

- [ ] Build a release action that works when PRs are merged:  It should release to GitHub Releases + Eddie's new release repository.  
- [ ] Go code should be released to go github releases with the right tags. 
- [ ] Update documentation / governance for the project on how releases are done - we should be saying that if the code is approved into the main branch, then releases can (and should automatically) happen.

CCC Releases should use the CalVer format as we've agreed before.  This should happen automatically on merge.

robmoffat:
Will review the automation decision in the steering committee

PR #1008, opened 2026-05-21
change(ci): Updated release pipeline

eddie-knight:
This change set allows us to publish releases to https://grc.store in addition to the GH releases page.

This is a first step — the PR was already getting big so I decided not to go through and bring every catalog into compliance with the latest schema. 

This disables the website updates due to the change in artifact shape. We should update that pipeline in a dedicated PR.

**Test run**: https://github.com/finos/common-cloud-controls/actions/runs/26257290484
**Result**: https://grc.store/search/finos-ccc/ccc.objstor.cp

I left a whole bunch of claude ultrareview comments inside the PR so that it is clear where we need to make follow-up PRs. I will be drafting those as soon as we merge this in.


eddie-knight:
I pinged @TheJuanAndOnly99 for guidance on the failing check. It can be safely ignored for now.

sshiells-scottlogic (approved):
Looks good as far as I can tell

schemas/capabilities-schema.json
```diff
@@ -18,8 +18,8 @@
               "properties": {
                 "reference-id": {
                   "type": "string",
-                  "pattern": "^CCC\\.Core.CP\\d{2,8}$",
-                  "description": "Global/common capability IDs"
+                  "pattern": "^CCC\\.Core\\..+$",
+                  "description": "Global/common capability IDs (CP## or F## during migration)"
                 },
```
robmoffat, line 23: can we not pull in these schema files from Gemara now?

.github/workflows/release.yml
```diff
@@ -1,103 +1,162 @@
...
+        env:
+          ARTIFACT: ${{ steps.artifact.outputs.file }}
+          GEMARA_VERSION: ${{ steps.artifact.outputs.gemara_version }}
+        run: |
+          git clone --depth 1 --branch "$GEMARA_VERSION" \
+            https://github.com/gemaraproj/gemara /tmp/gemara-spec
+          grcli validate -f "$ARTIFACT" --spec /tmp/gemara-spec
 
-      # Create a GitHub release
-      - name: Create GitHub Release
```
robmoffat, line 79: we need to create the GitHub release too

.github/workflows/release.yml
```diff
@@ -1,103 +1,162 @@
...
-        id: create_release
-        uses: actions/create-release@v1
-        with:
-          body_path: ./release_notes.md
-          tag_name: ${{ github.event.inputs.tag }}-rc
-          release_name: Release ${{ github.event.inputs.tag }}-rc
-          draft: false
-          prerelease: true
+      # NEW publishing step: push the typed catalog to grc.store.
+      - name: Publish to grc.store
```
robmoffat, line 131: Who or what owns grc.store?  Are we creating a new link to a proprietary service here?  

Is this infra owned by OpenSSF?

robmoffat (changes requested):
One further improvement I would like to see is that this should run automaticatically as well as workflow dispatch.

i.e. 
 - Build the catalogs.
 - Compare (perhaps just the IDs) to previous versions
 - If there is changed content, perform the release automatically.

That would mean we complete https://github.com/finos/common-cloud-controls/issues/993 too

schemas/capabilities-schema.json
eddie-knight: Yes we could achieve the vetting strictly through the CUE schema, but @sshiells-scottlogic requested that we persist the IDE support via jsonschema

.github/workflows/release.yml
eddie-knight: That was just moved down: https://github.com/finos/common-cloud-controls/pull/1008/changes#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R141-R146

Still there 👍

.github/workflows/release.yml
eddie-knight: That's all built and maintained by me at the moment. It'll unlock several capabilities in addition to what we have from the current GitHub release flow — especially valuable is the `grcli pull ...` command which will allow anyone to fetch official immutable release artifacts and vet provenance in a single command.

zigmax: @robmoffat Do you know if we can host that under the FINOS domain?

catalogs/core/ccc/groups.yaml
```diff
@@ -0,0 +1,89 @@
+# Canonical CCC group definitions — the single source of truth.
+#
+# Catalog entries (capabilities, controls, threats) reference a group by `id`.
+# At compile time the toolkit injects the full definition below into the
+# artifact's `groups`. Referencing a group id that is not defined here is a
+# compile error, which keeps every catalog's grouping consistent.
+groups:
```
zigmax, line 7: Maybe we could introduce two new groups focused on Container Orchestration and Agentic AI.

eddie-knight: Good idea. We can make a dedicated PR to address all of the groups in detail before the next release cycle

zigmax: Sounds good to me :)

zigmax (approved):
Looks good to me. The only open question is the grc.store domain, but we can discuss that during our weekly meeting.

smendis-scottlogic (approved):

robmoffat (approved):
removing my block - let's discuss the hosting in the all-hands.

PR #1010, opened 2026-05-26
change(ci): added group entries so that pre-release trials can commence

eddie-knight:
This adds groups that were generated by Claude, for the purpose of unblocking release pipeline trials. We plan to give each catalog a final review prior to each release, so these are not intended to be bulletproof at this time.

### DIFF
--- a/catalogs/ai-ml/gen-ai/capabilities.yaml
+++ b/catalogs/ai-ml/gen-ai/capabilities.yaml
@@ -32,6 +32,7 @@ imported-capabilities:
 
 capabilities:
   - id: CCC.GenAI.CP01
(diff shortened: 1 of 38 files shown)


smendis-scottlogic:
Where does the set of groups live to choose from?

zigmax (approved):

eddie-knight:
@smendis-scottlogic that's in a recently published entry:

https://github.com/finos/common-cloud-controls/blob/main/catalogs/core/ccc/groups.yaml

The groups aren't independently versioned but are part of the delivery-toolkit. So whatever is on the main branch at time of release will be inserted into the catalog's metadata.

PR #1013, opened 2026-05-28
change(ci): remove our customizations to semantic pr title requirements

eddie-knight:
We got creative a while back and tried customizing the PR title requirements, but the customized requirements never became habit or intuitive even for maintainers. This restores the normal semantic pr title requirements (ci, feat, docs, fix, etc)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,60 @@
+<!--
+=============================================================================
+  Thanks for contributing to Common Cloud Controls!
+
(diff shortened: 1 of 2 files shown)


sshiells-scottlogic (approved):

smendis-scottlogic (commented):
Now it has defaulted to use conventional commits format right?

Shouldn't we document how the titles should look like now since it still ask for specific types such as 
feat → new feature
fix → bug fix
docs → documentation
chore → maintenance
refactor → code restructuring
test → tests
ci → CI/CD changes

eddie-knight:
Good point @smendis-scottlogic — I added a PR template so that everyone is reminded of the reqs at the time it actually matters.

PR #1014, opened 2026-05-28
feat(controls): update core controls to Gemara v1.2.0

eddie-knight:
This looks like a 100% diff, but it's just a matter of un-indenting everything from `control-families` and putting the family names as the `group` in each control.


eddie-knight:
I'll merge this into the upcoming super-PR, per request from @zigmax

PR #1015, opened 2026-05-28
feat(controls): update genai controls to Gemara v1.2.0

eddie-knight:
This looks like a 100% diff, but it's just a matter of un-indenting everything from control-families and putting the family names as the group in each control.


eddie-knight:
@zigmax Would you prefer that I put all of the control catalog structure commits into one big PR? Since they're nearly impossible to review as a diff anyway, maybe we should just smash them through all together and do the final reviews before release like we keep talking about.

zigmax (approved):

zigmax:
@eddie-knight 1 big pr :)

eddie-knight:
Brilliant, I'll do the rest in that format

PR #1067, opened 2026-06-09
fix: resolve Gemara v1.2.0 catalog schema violations

eddie-knight:
## Summary

Five mechanical fixes that bring source catalogs into Gemara v1.2.0 schema compliance and let every affected target compile cleanly through `delivery-toolkit compile` + `grcli validate --spec v1.2.0`.

- `catalogs/compute/virtual-machines/threats.yaml` — add missing `group: Resource` to `CCC.VM.TH12`. Without it, `compute/virtual-machines/threats` fails to compile.
- `catalogs/management/auditlog/controls.yaml` — rewrite `CCC.AuditLog.CN10.AR02` from `"it should be denied by bucket policy"` to `"the request MUST be denied by bucket policy"` (RFC 2119 strength) and fix `"a audit"` → `"an audit"`.
- `catalogs/core/ccc/controls.yaml` and `catalogs/database/vector/controls.yaml` — remove three empty `guidelines: []` lists. CUE forbids empty lists outside `extends`.
- `catalogs/categories.yaml` — rename `CCC.RDMS` → `CCC.Database` (no other file references it as a *category id*; the catalog id namespace is unaffected) so the three `database/*` metadata files resolve. Also add new `CCC.Orchestration` and `CCC.Pipeline` categories referenced by `orchestration/k8s/metadata.yaml` and `orchestration/etl/metadata.yaml`.

## Verification

Compiled and validated every affected target after the changes:

- `grcli validate --spec gemara@v1.2.0` → OK for all 19 compiled artifacts under the touched targets (`compute/virtual-machines`, `core/ccc`, `database/{vector,relational,warehouse}`, `management/auditlog`, `orchestration/{etl,k8s}`).
- The pre-existing `compute/virtual-machines/controls.yaml` `"no native controls to compile"` failure is **out of scope** here; tracked in a follow-up PR that fixes the compile tool itself.

## Test plan

- [x] CI YAML schema check passes
- [x] CI lint check passes
- [x] Reviewer spot-checks `CCC.AuditLog.CN10.AR02` wording change is acceptable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1068, opened 2026-06-09
fix: scaffold missing controls and threats stubs for tracing, etl, k8s

eddie-knight:
## Summary

Three build targets — `catalogs/management/tracing/`, `catalogs/orchestration/etl/`, `catalogs/orchestration/k8s/` — ship a `metadata.yaml` and `capabilities.yaml` but are missing `controls.yaml` and `threats.yaml` entirely. This PR adds CCC.Core-import stubs to all three so the directory layout matches every other catalog and the targets are picked up by future tooling.

- Each new `threats.yaml` imports `CCC.Core.TH01`–`TH15` (same baseline used by `catalogs/compute/serverless-computing/threats.yaml`).
- Each new `controls.yaml` imports the same CCC.Core controls baseline used by `catalogs/compute/virtual-machines/controls.yaml`.

The stubs are imports-only. They will not compile through `delivery-toolkit compile` until #(follow-up PR) lands the "imports-only is OK" fix — the compile tool currently errors with `no native … to compile`. That's a known limitation that already affects 7 other source files; this PR just brings the missing targets up to the same baseline.

## Dependencies

This PR is independent of #1067 and the upcoming compile-tool fix. Once the compile-tool fix lands, every catalog in the repo will compile cleanly.

## Test plan

- [ ] CI YAML schema check passes for the new stub files
- [ ] CI lint check passes
- [ ] After the compile-tool fix lands, these targets produce a valid `ControlCatalog` and `ThreatCatalog` artifact

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/catalogs/management/tracing/controls.yaml
+++ b/catalogs/management/tracing/controls.yaml
@@ -1,0 +1,21 @@
+imported-controls:
+  - reference-id: CCC
+    entries:
+      - reference-id: CCC.Core.CN01
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN02
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN03
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN04
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN05
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN06
+        strength: 0 # Not yet specified
+      - reference-id: CCC.Core.CN07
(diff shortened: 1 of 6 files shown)


eddie-knight:
Closing — these stubs are imports-only, which is the same authoring tech debt that prevents the existing seven imports-only sources from compiling. Adding more imports-only files perpetuates the problem rather than fixing it. The three targets (`tracing`, `etl`, `k8s`) need native controls and threats authored before `controls.yaml` / `threats.yaml` should land. Re-opening will follow that content work.

PR #1125, opened 2026-06-19
fix: Fixing lint issues

robmoffat:
## Summary

Fixing the various linting and markdown checking issues on the website so we don't keep merging PRs with failing checks.

- #1095 

## Related issues

Based on `Eddie-knight:fix/refarch-ids`

PR #1098, opened 2026-06-16
fix: updated malformed ids in multi-agent-refarch

eddie-knight:
## Summary

Apparently I made zero effort to create real IDs for this catalog set. This corrects that issue and gets the automation working again.

_This PR did NOT involve a detailed manual review to qualify it for an official release. This is strictly focused on restoring functionality to CI and website builds._

## Related issues

#1095

## Checklist

- [x] PR title follows the Conventional Commits format (see the guide at the top of this template)
- [x] Tests / docs updated as needed

PR #1129, opened 2026-06-20
feat: move external mapping data to standalone mapping documents

eddie-knight:
## Summary

Previously the CI was giving a large number of errors due to the `external-mappings` field being used across the repo.

This extracts the external mappings into the high-fidelity Gemara `MappingDocument` artifacts. Mass removal to new files.

This does not yet address publication of the new artifact type.

## Related issues

part of #1125 
unblocks #1098

## Checklist

- [x] PR title follows the Conventional Commits format (see the guide at the top of this template)
- [x] Tests / docs updated as needed


robmoffat:
when we talked about this in the all-hands, you said you were going to write an explanation for why this is changing.  Is that still incoming?  As it stands, I don't understand why we are changing the catalog format again.

eddie-knight:
This isn't super new, it's just been deferred as a lower priority change compared to all the other bits.

We designed the mapping documents so that we can have higher fidelity references to external documents, and we can evolve those mappings over time without demanding a new release to an existing catalog.

robmoffat:
Ok @eddie-knight that makes sense, and now the conference is over, this is the right time for a big change like this.  

However I would like to see two things:

1) We need to make sure the other maintainers are onboard with this and happy with it happening.  So, discuss at the all-hands.  

2) We don't just change the catalogs on their own as this will break everything else - this needs to cover the skills Maxime and Sonali have been writing and the website stuff that @gbalysevaite and co have been working on... and anything else I've forgotten about! 

thanks

eddie-knight:
I don't see any reference to mappings in the skills, but I've gone ahead and added render logic for the mappings on the existing controls pages.

Screenshot 2026-07-07 at 9 13 42 PM

catalogs/core/core/mappings/threats-mitre-attack.yaml
```diff
@@ -0,0 +1,232 @@
...
+    remarks: Obfuscated Files or Information
+  - entry-id: T1485
+    remarks: Data Destruction
+  - entry-id: T1565
+    remarks: Data Manipulation
+  - entry-id: T1489
+    remarks: Service Stop
+  - entry-id: T1562.01
+    remarks: 'Impair Defenses: Downgrade Attack'
+  - entry-id: T1027
```
jpower432, line 204: Looks like this is a duplicate along with the next 4.

jpower432 (approved):
LGTM. Left one nit about mapping duplicates.

PR #1141, opened 2026-06-30
feat: generate sidebar menu titles from file system structure of catalogs

### LINKED ISSUE #1136: Catalogs Sidebar is Hard-Coded
## Bug Report

`CatalogSidebar.tsx` should be generating the list of catalogs from the contents of `website/src/data/ccc-releases`

AAloshine-scottlogic:
## Summary

sidebar menu is now generated by exploring the catalog filesystem for names, this adds in orchestration and app integration to the menu

## Related issues
closes #1136

### DIFF
--- a/website/src/content/catalogUtils.ts
+++ b/website/src/content/catalogUtils.ts
@@ -1,7 +1,17 @@
 import { SectionItem } from "./sections";
 
 // Prettify a slug segment or compound slug (e.g. "ai-ml" → "AI/ML", "gen-ai" → "Gen AI").
+const SLUG_DISPLAY_NAMES: Record<string, string> = {
+  mlde: "MLDE",
+  ccc: "CCC",
+  devtools: "DevTools",
+  batchproc: "Batch Processing",
+  auditlog: "Audit Log",
+  loadbalancer: "Load Balancer",
+};
+
 export function prettifySegment(s: string): string {
+  if (SLUG_DISPLAY_NAMES[s]) return SLUG_DISPLAY_NAMES[s];
   const acronyms = new Set(["ai", "ml", "iam", "vpc", "etl", "k8s", "sdk"]);
   const parts = s.split("-");
   if (parts.every((p) => acronyms.has(p))) {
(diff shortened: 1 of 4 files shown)


gbalysevaite:
Looks good to me, @robmoffat can you check if you're happy with the sidebar names, for example K8S for Kubernetes. Anna has added some SLUG_DISPLAY_NAMES, so more can be added there as needed

gbalysevaite (approved):

website/src/content/catalogUtils.ts
```diff
@@ -1,7 +1,17 @@
 import { SectionItem } from "./sections";
 
 // Prettify a slug segment or compound slug (e.g. "ai-ml" → "AI/ML", "gen-ai" → "Gen AI").
+const SLUG_DISPLAY_NAMES: Record<string, string> = {
```
robmoffat: Ok, I don't think this Is the way.  

Looking at the metadata file we have these two fields `id` and `title`:

```
 id: CCC.MLDE
  title: CCC Machine Learning Development Environment
```
 
or

```
id: CCC.MARefArc
  title: Multi-Agent Reference Architecture
```

or

```
id: CCC.AuditLog
  title: CCC Audit Logging
```

or

```
id: CCC.BatchProc
  title: CCC Batch Processing
```

I don't think it makes sense to maintain a third mapping somewhere else for this.

I get that we are trying to keep it short, so would it not be better to construct the title based on the ID?

e.g. 

1.  Strip the CCC.
2. Turn CamelCase into Title Case

Alternatively, we could construct a different menu based around titles, or have a popover or something.  WDYT?

gbalysevaite: I understand your want to have only one point of truth, that would be best. However, if we use the ID as you suggested 'CCC.MARefArc' would show as 'MA Ref Arc' which is not understandable for me. 

@AAloshine-scottlogic can you make the sidebar using titles instead? Just strip the 'CCC', let's see how it looks

AAloshine-scottlogic:
for some reason the preview isn't showing the additional catalogs (app integration/ orchestration) but they are showing up locally 
image

EDIT: this is now showing up in the deployed preview

website/src/content/catalogUtils.ts
robmoffat: good call @gbalysevaite !

robmoffat (approved):

gbalysevaite:
@AAloshine-scottlogic feel free to merge these changes and we'll improve further

